Issue #152, opened 2015-11-13
derivates to construct constraints


nikohansen:
In order to construct constrained problems in a controlled way, we would like the underlying objective functions to deliver their derivative. 

Proposal: a function like

```coco_get_derivative(coco_problem_t *problem, double *x, double *dx)```

or 

```coco_evaluate_derivative(coco_problem_t *problem, double *x, double *dx)```

is implemented which outputs in `dx` the derivate at point `x`. Each `f_...` file that implements the underlying function used with `coco_evaluate_function` also implements the underlying function used with `coco_evaluate_derivative`, in case it is to be used for a testbed with constraints.

brockho:
This issue seems to be solved with the `development-constraints` branch, or am I wrong?

phrsampaio:
Yes, it is solved there.

PR #205, opened 2015-12-07
Coco constraints

phrsampaio:
Modifications in the code related to Issue #152 and to the suite of constrained problems.


brockho:
I think that evaluate_gradient should not have the gradient as first and the search point as second parameter because it is a) highly confusing and b) not consistent with e.g. evaluate_function. The other way round seems more natural to me.

phrsampaio:
The corrected file is in my fork. All the functions related to the gradient in the other files have the search point followed by the gradient though, so I probably forgot to modify it in f_sphere.c.

PR #301, opened 2016-01-14
Fix for 2 or more bi-objective algorithms.

dtusar:


nikohansen:
I don't understand why the post-processing can be confused by several objectives. To my understanding it only sees a single indicator value?

dtusar:
In the bi-objective case we don't have the tdat files so some fields in the dataset object are not set. This fix is for one of the errors from the email Dimo sent yesterday.

Issue #400, opened 2016-01-19
Move log_level out of the observer options


ttusar:
`log_level` is not really connected to the observer and should be accessible in some other way. Maybe we should make a `coco_set_log_level()` function in order to be able to change it...

ttusar:
Done! 

The `"log_level: string"` option in the observer options will be ignored now, therefore the other languages need to implement an interface to the `coco_set_log_level()` function if they wish to set the `log_level` to some other value than the default (`info`). I just did this for Java (it's not on the development branch yet, but will be shortly). 

TODO: Matlab/Octave and Python

BTW, do we also need a `coco_get_log_level()` function?

ttusar:
Issues #433 and #434 have now been created for this, therefore I'm closing this issue.

nikohansen:
"_BTW, do we also need a coco_get_log_level() function?_" 

I just stumbled of the use case where I want to change the level temporarily. So yes, but in this case it would be sufficient for the setter to return the old level. 

Input "" could then serve as indicator to set nothing but return the current level. 

Other unknown input values should issue a coco_warning. 

So, no, but...

ttusar:
Isn't it cleaner to just have a `coco_get_log_level()` function?
I will take care of the warning.

nikohansen:
I believe the above is a common design pattern. It allows for
```
old_level = coco_set_log_level(new_level) // enter
dirty work work work
coco_set_log_level(old_level)  // exit
```
instead of 
```
old_level = coco_get_log_level()  // prepare
coco_set_log_level(new_level)  // enter
dirty work work work
coco_set_log_level(old_level)  // exit
```
which is considerably more ugly. 

One could have both...

ttusar:
I see how it can be "easier" to use, but I still think one expects the getter to return some value and the setter to set it, not a setter to get it and set it...

ttusar:
What if I provide the two "clean" functions in C and then you do your magic in Python within the interface?

nikohansen:
Another advantage, I admittedly did not mention, for having only one function is that the interface remains (c)leaner. Having getter and setter functions is rather needless clutter. And we might come up with more variables we like to be able to manipulate in future. 

Looking at the `build/matlab` folder is rather ugly. That's of course also the problem of Matlab, but keeping an interface to a small number of functions is, I believe, a value in itself.

I let you decide anyway. 

[...]

PR #431, opened 2016-01-20
Changed log_level handling, output of suite iteration and coco-doc/C

ttusar:


nikohansen:
We now need to propagate the new function `coco_set_log_level` into all language-dependent interfaces, right to be able to use it?

ttusar:
Sure, see issue #400

PR #477, opened 2016-01-22
Development

asmaatamna:
Changed .c files in matlab directory that were causing warnings under matlab and octave


brockho:
pull request seems fine from my side - these were exactly the occurrences of `mwSize` that I have missed

Issue #325, opened 2016-01-15
logger output at prescribed times


nikohansen:
The logger should write additionally a output line for fixed number of f-evaluations. 
For example at `dimension * [1, 2, 5, 10, 20, 50, 100, 200, ...]`. 
This allows, for example, for rigorous statistical testing up to a given budget, when all compared algorithms surpassed this budget.

ttusar:
Is up to 200 enough?

nikohansen:
no! the ellipsis ... was there to suggest that this sequence goes to infinity, it should preferably dynamically generated, e.g. from the sequence `small_factors = [1, 2, 5]` like
```
if (idx_small_factor < 2) {
    idx_small_factor += 1
    factor *= small_factors[idx_small_factor] / small_factors[idx_small_factors - 1]
} else if (idx_small_factor == 2) {
    idx_small_factor = 0
    factor *= 10 / small_factors[2]
} else 
    raise ValueError() ;-)
```

ttusar:
Should this be output in the `*.dat` files or the `*.tdat` files (currently non existing in the biobj logger)?

nikohansen:
As additional line in the `.dat` file (given a line has not been written at this time step).

ttusar:
Done for the biobjective logger. Should the issue stay opened for some other logger or can we close it?

nikohansen:
Let's have this for the `bbob` observer as well.

ttusar:
@oaelhara You can use the `coco_observer_evaluation_to_log(size_t number_of_evaluations, size_t dimension)` function in `coco_observer.c` to check if the current evaluation number matches the predefined values. [It's now being tested on the test1 branch, but it should be moved to development soon.]

oaelhara:
Thanks for the pointer @ttusar.
I'm not sure I understand why it should be in the .dat files not the .tdat files when the latter are the ones that contain the number of function evaluations aligned data. As I see it, we want additional, dimension dependent, nbfevals triggers which don't belong in the function value aligned data files (.dat) but in the .tdat files.

ttusar:
The `dat` vs. `tdat` discussion was for the biobjective logger, which does not yet have a `tdat` file (and won't need one as it seems). I think for the bbob logger it makes more sense to put it in the `tdat` files (I'm sure @nikohansen will correct me if I'm wrong :)).

oaelhara:
Implemented for the bbob_logger, writes to the .tdat files

[...]

PR #496, opened 2016-01-25
Reading the evalution based data from .dat files.

dtusar:


oaelhara:
So this means in #325 the data logging at fixed f-evals*dim should now be done (also/just?) to the .dat files?

dtusar:
It is not decided yet.

nikohansen:
If they are logged in the `.tdat` files already, leave it like that for the time being.

PR #649, opened 2016-02-15
Development

asmaatamna:
transform_vars_*.c that don't update best_parameter now generate a warning if best_parameter is different from zero.


nikohansen:
To me it looks OK. Can someone else have a look as well?

ttusar:
You might want to consider putting the while loop in a function (since it's been used so often in different places), but other than that, I don't see any problems.

nikohansen:
So should we merge it? The longer we wait, the more likely we (i) will see conflicts and (ii) the introduced bugs will show up in relevant data.

ttusar:
OK from my side.

Issue #707, opened 2016-02-26
documentation: parallelization of the suite


Ulfgard:
Hi,

I am wondering whether it is possible to parallelize the functions of the suite (e.g. going through each problem and starting a thread for each). This is because currently evaluation can take a long time and 94% of my cores are idle.

I am looking at this simple loop:

```
coco_suite_t* suite = coco_suite("bbob-biobj", "year: 2016", "dimensions: 5");
coco_observer_t* observer = coco_observer("bbob-biobj", "...");

coco_problem_t* problem;
while ((problem = coco_suite_get_next_problem(suite, observer)) != NULL) {
    //this while loop should be run in parallel
}
```

the problem is, that this is not a "nice" loop for easy parallelization with OpenMP. for that a for loop over all problem numbers would be much easier.

I am currently thinking about the following solution in C++:

```
//first query all problems
std::vector<coco_problem_t*> problems;
coco_problem_t* problem;
while ((problem = coco_suite_get_next_problem(suite, observer)) != NULL) {
    problems.push_back(problem);
}
//run them in parallel
PARALLEL_FOR(std::size_t i = 0; problems.size(); ++i){
    //...
}
```

But the documentation does not state whether the observer is thread safe, whether several independent observers are possible or whether the problem instances actually exist after the next call of of coco_suite_get_next_problem.

ttusar:
Your are right, the documentation doesn't address this, we need to expand it. 

Your example won't work because the problem instance returned by `coco_suite_get_next_problem` is available only up to the next call to the same function. Your while loop can be substituted by this (note it's untested):
```C
for (std::size_t index = 0; index < coco_suite_get_number_of_problems(suite); index++) {
    problem = coco_suite_get_problem(suite, index);
    problems.push_back(problem);
}
```
However, the big question is what to do with the observers. The problems above are not observed and you need as many observers as you have parallel threads (this should work fine). I'm not sure what `PARALLEL_FOR` does exactly (I'm not familiar with OpenMP), but you need to wrap each problem with an observer using `coco_problem_add_observer()` before you run your algorithm on it. 

In the end you will also need to free all problems and all observers. It is enough if you free each problem using `coco_problem_free(problem)`, i.e., you don't need to explicitly remove the observer first using `coco_problem_remove_observer()`.

nikohansen:
If we create a single observer instance for each problem, this is supposed to work, at least in my mind. For this very reason we tried to get rid of global/static variables in the observer implementation.

[...]

PR #726, opened 2016-02-29
fixed issue in observer

Ulfgard:
When the observer should log the indicator in the info file, it checks whether the current function is different from the previous function observed. It does not check whether the dimensionality is different. this breaks cases where the same function is evaluated in different dimensions.

problem came up in #707


nikohansen:
Nice catch.

ttusar:
Thanks!

PR #841, opened 2016-03-16
Update hypervolume reference values

ttusar:


ttusar:
Will make a new one with additional 10D and 20D data

PR #855, opened 2016-03-18
Update hypervolume

ttusar:


brockho:
Comparing random search wrt the old and the new HV reference values gives almost no difference because the number of targets, we record, is large (100 per order of magnitude) and the improvement of the reference set's hypervolume is relatively small.

With the new selection of displayed targets, however, i.e. 33 targets including 0 and some negative vs. the previous 51 positive targets, we see a difference in the ECDF plots: for small dimensions, the plots are lower (but of the same shape; as expected) and for the larger dimensions, the displayed performance is similar or even better than before. In particular, the step at the first evaluation becomes more pronounced than before as we have less targets overall.

Hence, I am in favor of merging the new hypervolume values of the improved reference sets but also advocate to (almost) double the targets displayed to something like `{-10^{-4}, -10^{-4.1}, -10^{-4.2}, \ldots, -10^{-4.8}, -10^{-4.9}, -10^{-5}, 0, 10^{-5}, 10^{-4.9}, 10^{-4.8}, \dots, 10^{-0.1}, 10^0}`.

pprldmany_f002-comp

brockho:
EDIT: For RANDOMSEARCH, still a few ECDFs look "strange" with the updated reference hypervolume values:
* f12 and f15 (where the 3-D curve is better than the 2-D one)
* f20 (where 3-D, 5-D, and 10-D almost look the same)
* f23 (where 2-D and 3-D are almost the same)
* f24, f46 and f50 (where 10-D and 20-D are almost the same
* f44 (where 3-D and 5-D are almost the same before the cross)
* f48 (where 10-D is better than 5-D) and
* f53 (where 3-D is better than 2-D in the range of 1.5D-3.5D funevals)

ttusar:
@brockho I would like to have a closer look at what is going on. Can you upload these results by random search? Also, was this strange behavior seen for a particular instance or an aggregation over all instances for the listed problems?

brockho:
Here are the data, I was looking at:

[ppdataRS-NEW-HV.zip](https://github.com/numbbo/coco/files/180714/ppdataRS-NEW-HV.zip)

Unfortunately, some of the 40-D results are missing, because my laptop decided to reboot with a Windows update in between :-)  The statements were all made on the ECDF plots per function (i.e. aggregated/bootstrapped over all instances, see the file `RS_on_bbob-biobj-3e4funevalsNEWHV/pprldmany-single-functions/pprldmany.html`).

nikohansen:
> With the new selection of displayed targets, however, i.e. 33 targets including 0 and some negative vs.

33 targets was a "bug", I indeed wanted to suggest to add to the 51 positive targets + zero the 6 targets between -10^-4 and -10^-5, pretty much as define above. This means that 6/58 targets (10%) are smaller than zero.

That the display does not change is indeed strange.

PR #873, opened 2016-03-22
Development

tobiaswagner:


brockho:
Unfortunately, the pull request has some rather serious issues:
* the SMS-EMOA code does not run under Octave on my Windows machine due to an incompatible `version` command
* the NSGA-II code does not run either under Octave for me:
```
>> run_gamultiobj_on_bbob_biobj
COCO INFO: Results will be output to folder exdata\gamultiobj_on_bbob-biobj
"gamultiobj on the bbob-biobj suite...

COCO INFO: 22.03.16 20:54:55, d=2, running: f01.Optimizing bbob-biobj_f01_i01_d02
error: @gamultiobj: no function and no method found
error: called from
    my_gamultiobj at line 3 column 9
    run_gamultiobj_on_bbob_biobj at line 28 column 9
error: evaluating argument list element number 1
error: called from
    my_gamultiobj at line 3 column 9
    run_gamultiobj_on_bbob_biobj at line 28 column 9
```

After some testing and a first fix of the `version` issue above, @tobiaswagner and me found out that on both our machines (under Octave), the SMS-EMOA crashes with segmentation faults that seem to come from the external `paretofront` function. Interesting that, so far, this has not been occurring with the current code.

I therefore suggest, we wait with the integration and do not include the new/updated examples into the next release.

loshchil:
sometimes this sort of errors happen when a function is not linked to the
code, e.g., if one call feval('myfunc', ...) in the code, then when mcc
builds a standalone application it does not link myfunc.m as it is not
called directly but by feval and string. A solution to this is to force the
inclusion of the myfunc.m . Sorry if this has nothing to do with the error
you found.

PR #951, opened 2016-03-31
Started reviewing the general performance assessment.

ttusar:
It seems that my text editor has deleted some spaces at the end of the lines by itself, so it looks as if more changes were done. I've only reviewed the document up to "Missing values".


brockho:
I am fine with accepting this pull request. All changes actually made are fine by me, the suggested changes in the comments should be discussed (but after the merge I suggest).

PR #960, opened 2016-04-05
corrected bug in pptables which occurred during nightly build

brockho:


dtusar:
Hi Dimo,

I'm sorry that I didn't send you an email, but I thought that you would 
anyway get the github email. I already fixed this in the morning so you 
didn't have to do it.

Best,
Dejan

PR #1035, opened 2016-05-19
Updates for the suite of constrained problems

phrsampaio:


brockho:
Here my first bits, before running the tests:

`code-experiments/build/c/example_experiment.c`:
* All changes needed?
* BUDGET_MULTIPLIER should be back to 2
* probably one general random search that can run on any available testbed?

`code-experiments/src/coco_internal.h`
* why is `typedef void (*coco_initial_solution_function_t)(const coco_problem_t *problem, double *y);` gone?


`code-experiments/src/coco_problem.c`
* many indentations screwed
* introduced misspelling in comments like
 replacing `@note Can be used to prevent unnecessary burning of CPU time.` with `@note Can be used to prevent unnessary burning of CPU time.`
* why are several asserts removed?

`code-experiments/src/coco_utilities.c`
* comment `@brief Normalizes vector x, then scales it by a factor alpha.` seems wrong

`code-experiments/src/f_bent_cigar.c`
`code-experiments/src/f_different_powers.c`
`code-experiments/src/f_discus.c`
`code-experiments/src/f_ellipsoid.c`
`code-experiments/src/f_linear_slope.c`
`code-experiments/src/f_rastrigin.c`
`code-experiments/src/f_sphere.c`
* again some indentation changes

`code-experiments/src/logger_bbob.c`
* why removing existing comments?
* why changing `lu` to `ld`?
* again: why removing asserts?
* why is `logger_bbob->written_last_eval` changed?

`code-experiments/src/transform_obj_shift.c`
* indentation problem

`code-postprocessing/bbob_pproc/testbedsettings.py`
* `get_benchmarks_short_infos` seems incorrect, because not consistent with `bbob` testbedsettings

ttusar:
The naming of files, functions and types should follow the "coco naming convention". For example, all files related to constraints could start with a `c_` in the same way as all files related to functions start with `f_`. Please see http://numbbo.github.io/coco-doc/C/globals.html to get a feeling of how this is currently done.

phrsampaio:
brockho commented
...
code-experiments/src/coco_internal.h

> why is typedef void (*coco_initial_solution_function_t)(const coco_problem_t *problem, double *y); gone?

Since we know the initial solution by the construction of the constrained problems, it is simpler and more practical to consider initial_solution as a vector in the coco_problem_s struct, just like the best_parameter, instead of a function. The value of initial_solution is then updated each time the problem is transformed as it is done with best_parameter.

code-experiments/src/coco_problem.c

> many indentations screwed

Many indentations were already differing from one another due to the use of tabs, but I will check and correct what is misplaced.

> why are several asserts removed?

Which ones? The asserts were just moved by a few lines in the respective functions due to the adaptations made for the constraints. The only one removed that I see was the one at line 430 of coco_problem.c, which will be corrected in the next commits.

code-experiments/src/logger_bbob.c

> why removing existing comments?

I think because that some variables had been removed due to many tests and so were the comments in the same line. After putting the variables back, the comments were still gone. This will be corrected.

> why changing lu to ld?

That is strange, since I did not make these changes. It seems an older version of the file, which should be updated after the merge. It will be corrected.

> again: why removing asserts?

Same answer as for the one related to coco_problem.c.

> why is logger_bbob->written_last_eval changed?

It is used in logger_bbob_free() to write (or not) the information related to the last iteration.

ttusar:
> >    many indentations screwed
> 
> Many indentations were already differing from one another due to the use of tabs, but I will check and correct what is misplaced.

The wrong indentations were probably my fault (stemming from changing computers some time ago), so I think @phrsampaio was right to correct them.

phrsampaio:
Regarding the only assert removed, I must have removed it because that there are now problems with no objective functions, only with constraints. Therefore, the function shouldn't throw an error when the number of objectives is zero. So it seems to make sense to remove it.

nikohansen:
> Since we know the initial solution by the construction of the constrained problems, it is simpler and more practical to consider initial_solution as a vector in the coco_problem_s struct, just like the best_parameter, instead of a function.

I might be wrong in this case, but in general one important reason why to "hide" getting or setting a value, like initial solution, behind a function is to have a clear separation between the interface and the implementation. This is done, because the implementation can and should change more often than an interface. Only this allows, for example, to add asserts when the value is demanded or even generate the value dynamically if need be in future _without breaking any past code_. If you have been programming in `Java`, this is why we see all these ugly attribute getter and setter functions.

phrsampaio:
The same rationale for defining best_parameter as a vector may be considered for initial_solution. Both points are known a priori, thus being set in the "constructor" functions. The vector initial_solution is set when constructing a linear constraint coco problem; its values are then copied to the final constrained problem when the latter is constructed via coco_problem_stacked_allocate(). Besides, whether the initial solution will be calculated differently in the future for different problems or not does not change the fact that it will be known by the time of building the problem, even if it has to call a function that would compute it. At last, the current way seems simpler than running through the stack of problems, having to distinguish all the transformations applied, to fetch all the required information to finally compute a point that was known from the very beginning.

brockho:
The difference between best_parameter and initial_solution is that the latter is supposed to be available to the algorithm while the former is definitely not. Niko's remark on breaking any past code relates to this, I guess: we don't want that previous code is not running anymore when internal things change. Hence, we should keep getting the initial solution as a function that is available to the optimization algorithm.

phrsampaio:
> The difference between best_parameter and initial_solution is that the latter is supposed to be available to the algorithm while the former is definitely not.

If you look at coco_problem.c, the function coco_problem_get_initial_solution() is still there, but now it returns the vector problem->initial_solution. So the algorithm has access to it.

> Niko's remark on breaking any past code relates to this, I guess: we don't want that previous code is not running anymore when internal things change. 

I might be missing something, but I don't see why this approach would not run when internal things change while having initial_solution as a function would not be affected. Would you have a possible scenario in mind? As I said before, the way that an initial solution is computed is by no means a concern in this approach. It is a set while building a constraint coco problem; whether it calls another function to compute it or not will depend on the type of the constraint only.

nikohansen:
> I might be missing something, but I don't see why this approach would not run when internal things change while having initial_solution as a function would not be affected.

Right, the function would be affected and would need to be changed. The point is that _only_ the function would need to be changed, while everything that _uses_ the function would not break. That makes a huge difference. The other point is that the function can be changed to influence how the initial solution is accessed, for example by adding asserts etc. These are two important reasons why to hide access to a field behind a function.

phrsampaio:
> The point is that only the function would need to be changed, while everything that uses the function would not break. That makes a huge difference. The other point is that the function can be changed to influence how the initial solution is accessed, for example by adding asserts etc. These are two important reasons why to hide access to a field behind a function is quite a useful concept. (Again, I didn't check the specifics involved in this case).

I agree. But maybe we are mixing the roles of coco_problem_get_initial_solution() and initial_solution as a function, which do the internal calculation of the point and that strongly depends on each constraint type. Your arguments seem to apply to the former. In this case, the approach of having initial_solution as a vector satisfy your two reasons: anything using coco_problem_get_initial_solution() would not break if internal changes are made and the field related to the initial_solution is kept well hidden. On the other hand, if internal changes were made at the sheer level of calculation of the initial solution, which again strongly depends on the constraint type, it seems natural that such changes be made at the time the constraint coco problem is built, which would not affect anything else in higher levels.

Issue #1026, opened 2016-05-14
Figures in "Average number of f-evaluations to reach target" are quite strange in large scale.


NDManh:
Hi, please find enclosed the result of post-processing for dataset in large scale (Random Search  for dimension from 20 to 1280). With the BUDGET = 1.0e4, it's hard to reach the easiest target, say 1.0e1, for even easy functions such as sphere, ellipsoid,.... In this case, I and Dimo have found some strange figures in "Average number of f-evaluations to reach target" (click on link `Average runtime versus dimension for selected targets`).
[largescale51.zip](https://github.com/numbbo/coco/files/264739/largescale51.zip)

nikohansen:
It looks like the budget is 1e4, not the budget multiplier. 

Generally it's not a big surprise to see nothing with random search in large dimension. Can you cross check with the random search papers / data which we have already in 20- and 40-D? That's one reason why we have overlapping dimensions. 

What is strange is that there is not even a maximum budget shown for example on the sphere or attractive sector. Did you check that the data look fine?

brockho:
It's not surprising that we see no ERT marker for RANDOMSEARCH for many functions (cp. http://sci2s.ugr.es/sites/default/files/files/TematicWebSites/EAMHCO/contributionsGECCO09/p2479-auger.pdf), but the marker for the final budget we should always see. Hence, I consider this a bug (except if the data is corrupt, but I doubt this).

NDManh:
`It looks like the budget is 1e4, not the budget multiplier.`
You're right, sorry since in this case I run the experiment in C in stead of Matlab and I am confused about this parameter. 

`Did you check that the data look fine?`
I might be wrong but I think that the data is normal. You can find the data set I post below. 

I'll carefully study the second term (cross check with the random search papers / data) and report soon.
[largescale51.zip](https://github.com/numbbo/coco/files/265135/largescale51.zip)

NDManh:
There is marker for the final budget with data of random search which I get from http://coco.gforge.inria.fr/doku.php?id=bbob-2009-algorithms. 

[ppdata.zip](https://github.com/numbbo/coco/files/265209/ppdata.zip)

NDManh:
For figures in `largescale51`: it seems that when 0 is on top-left corner we don't see the marker of maximum budget, i.e, there is nothing below 0. The other cases, we see these marker as usual.

NDManh:
I think I know where this bug comes from, and can solve it but I need your advice to make it optimal.
[...]

PR #1176, opened 2016-09-02
Fix the issue of nothing to plot #1026 and big merge with branch development 

NDManh:


code-postprocessing/bbob_pproc/ppfigdim.py
```diff
@@ -480,6 +483,13 @@ def plot(dsList, valuesOfInterest=None, styles=styles):
             res.extend(plt.plot(tmp[:, 0], maxevals / tmp[:, 0]**ynormalize_by_dimension,
                        color=styles[len(valuesOfInterest) - 1]['color'],
                        ls='', marker='x', markersize=20))
+            if no_target_reached:
+                ylim_maxevals = max(maxevals / tmp[:, 0]**ynormalize_by_dimension)
+                ylim = (ylim[0], math.ceil(ylim_maxevals))
+                plt.annotate('no target reached', xy=(0.5, 0.5), xycoords='axes fraction', fontsize=14,
+                            horizontalalignment='center', verticalalignment='center')
+            else:
+                ylim = (ylim[0], math.ceil(ylim[1]))
```
nikohansen: looking at the code more carefully, it looks like `ppfig.discretize_limits()` is meant to do this job (later) in `beautify()`. Hence `math.ceil` is not needed here or in line 488 (and we should see the same result without).

nikohansen:
It looks like the second commit should not be necessary. It shouldn't be too difficult to avoid getting this large change-set?

code-postprocessing/bbob_pproc/ppfigdim.py
NDManh: I have tried to change the code as you suggested, that is without using `math.ceil`:
```
if 1 < 3: # maxevals
            ylim = plt.ylim()
            res.extend(plt.plot(tmp[:, 0], maxevals / tmp[:, 0]**ynormalize_by_dimension,
                       color=styles[len(valuesOfInterest) - 1]['color'],
                       ls='', marker='x', markersize=20))
            if no_target_reached:
                ylim_maxevals = max(maxevals / tmp[:, 0]**ynormalize_by_dimension)
                ylim = (ylim[0], ylim_maxevals)
                plt.annotate('no target reached', xy=(0.5, 0.5), xycoords='axes fraction', fontsize=14,
                            horizontalalignment='center', verticalalignment='center')
            plt.ylim(ylim)
```
However, the obtained result is not the same, for example : 
- With the current code: 
[ppfigdim_f007.pdf](https://github.com/numbbo/coco/files/467257/ppfigdim_f007.pdf)
[ppfigdim_f023.pdf](https://github.com/numbbo/coco/files/467258/ppfigdim_f023.pdf)
- With the revised code (without using `math.ceil`):  
[ppfigdim_f007.pdf](https://github.com/numbbo/coco/files/467260/ppfigdim_f007.pdf)
[ppfigdim_f023.pdf](https://github.com/numbbo/coco/files/467262/ppfigdim_f023.pdf)

nikohansen:
I would like to understand why beautify doesn't do the job before to move on.

nikohansen:
I suggest also to first checkout the latest version of the branch and then add the small change, to un-clutter the commit (which has now 180 files changed).

NDManh:
> I would like to understand why beautify doesn't do the job before to move on.

1) Firstly, let's talk about the command `ylim = plt.ylim()` (line 482 in `ppfigdim.py`). This command gives different values to `ylim` in different runs of the post-processing on the same data set. For instance, I've tried to test on the data of the function `f23`, and obtained: 
`ylim = (0.0, 1.2000000000000002)` or `ylim = (0.0, 0.40000000000000002)` or `ylim = (0.0, 0.70000000000000007)`.

2) While in the case `ylim = (0.0, 1.2000000000000002)`, the result is the same for the code with or without using `math.ceil`; the difference appears for the other cases.

With the current code (using `math.ceil`), due to the command `ylim = (ylim[0], math.ceil(ylim[1]))` , the value `ylim = (0.0, 0.40000000000000002)` or `ylim = (0.0, 0.70000000000000007)` becomes `ylim = (0.0, 1.0)`, then `ppfig.discretize_limits()` in `beautify()` transforms it to `(ymin_new, ymax_new) = (0.6309573444801932, 3.000003)`. That is why we see the figure like  [ppfigdim_f023.pdf](https://github.com/numbbo/coco/files/475007/ppfigdim_f023.pdf)

In the case we do not use `math.ceil`, then `ppfig.discretize_limits()` in beautify() will transform `ylim = (0.0, 0.70000000000000007)` to `(ymin_new, ymax_new) = (0.6309573444801932, 1.0000009999999999)`, thus the figure looks like [ppfigdim_f023.pdf](https://github.com/numbbo/coco/files/475041/ppfigdim_f023.pdf)

nikohansen:
OK, thanks! Without `ceil`. If we don't like the result without `ceil`, we should rather change `discretize_limits`.

NDManh:
In fact, we do not adapt `max-evals` in this case, that is there is something rather than nothing (you can see the figure in the case using `math.ceil`). But it seems to be hidden, maybe it is because the `ylim` is too small?

PR #1181, opened 2016-09-12
replaced pickled IPOP-CMA-ES and RANDOMSEARCH data by raw data...

brockho:
...of IPOP-CMA-ES and MCS (only affecting a single doctest at the moment.


brockho:
There was a problem in pushing my latest commit to my fork before the pull-request. I'll try in a few minutes again.

PR #1197, opened 2016-09-26
Updating the devel-LS branch

NDManh:


nikohansen:
@oaelhara, could you confirm?

oaelhara:
The devel-LS branch is already over 200 commits behind the master branch and the merge request seems up to date with the development branch (well up to September 13th, a few commits behind) so I think it would be a good idea to have a fresher version of the large-scale code that was implemented more easily accessible (outside of our respective branches).

In addition, last time I checked, there are no "major" problems in the current large-scale code (I think most CI tests passed) but it should still be noted that it's probably not yet ready to be merged with the development branch.

nikohansen:
How do we proceed then?

nikohansen:
@oaelhara what are you suggesting?

oaelhara:
To merge with the devel-LS branch for now (accept the pull request).
We'll see about the merge with the development later.

Issue #1144, opened 2016-08-12
A postprocessing test should also use the latest data


ttusar:
The issue #1141 has shown that postprocessing needs to be run also on the latest data (I mistakenly thought this was already the case). At least in the nightly build, if this takes too long to do every time.

nikohansen:
For the next time I also suggest that we do not rely on integration testing to see whether a data format change breaks the data processing chain. I indeed doubt that this is what integration testing is meant to be for.

ttusar:
It goes without saying (now).

brockho:
Now that also the nightly build is passing again all tests,  the issue can be closed.

PR #1202, opened 2016-09-27
integration test for newly generated data (addressing #1144)

brockho:
To this end, the `example_experiment.py` was changed to handle experiments on other than the `bbob` test suite.


brockho:
an improved pull-request will follow soon

PR #1204, opened 2016-09-28
Updated scaling for the constraints

phrsampaio:


code-experiments/src/c_linear.c
```diff
@@ -310,22 +317,31 @@ static coco_problem_t *c_linear_single_cons_bbob_problem_allocate(const size_t f
...
+   * and only U_i is drawn for each constraint individually. 
+   * The random number 10**U[0,1] is given by the variable 'factor1' 
+   * while the random number 10**U_i[0,2] is calculated below and 
+   * stored as 'factor2'. (The exception is when the number of
+   * constraints is n+1, in which case 'factor2' defines a random
+   * number 10**U_i[0,1])
+   */
+     
+  if (number_of_linear_constraints == dimension + 1)
+    exp2 = rand() / (RAND_MAX / (1 + 1) + 1);
```
nikohansen: why don't you use the generator `coco_random_new`?

code-experiments/src/c_linear.c
```diff
@@ -310,22 +317,31 @@ static coco_problem_t *c_linear_single_cons_bbob_problem_allocate(const size_t f
...
+  else 
+    exp2 = rand() / (RAND_MAX / (2 + 1) + 1);
```
nikohansen: AFAICS, this produces number between zero and, given `RAND_MAX>>3`, close to 3 (specifically `3R/(R+3)`), which I believe was not intended.

code-experiments/src/c_linear.c
phrsampaio: I had problems using coco_random_normal() to generate the gradients a long time ago. I can't quite recall what happened, but, after having generated a certain number of pseudorandom numbers, a segmentation fault occurred. The lack of comments in the functions in coco_random.c makes the debugging and its use rather difficult. Of course, anyone that knows coco_random.c well can change c_linear.c in order to use its functions.

code-experiments/src/c_linear.c
phrsampaio: It seems that one way for having `U[M,N]` would be:
1) Declaring `exp2` as an `int`
2) `exp2 = (rand() / (RAND_MAX + 1.0)) * (N - M + 1))`

Because of `1.0`, `(rand() / (RAND_MAX + 1.0))` is a double between 0.0 (inclusive) and  0.9999 (to four decimal places). Assume that `M = 0`. `(rand() / (RAND_MAX + 1.0)) * (N - M + 1))` will then be between 0.0 and (`N` + 0.9999). Since `exp2` is an `int`, the fraction is discarded and it assumes a value between 0.0 and `N`.

phrsampaio: Indeed.

code-experiments/src/c_linear.c
nikohansen: Do we have an issue for this bug? 

Unfortunately any change of the "random" number generator can only happen before the code goes into the release, because the instances should never change again afterwards. 

On a similar note, we probably have no guaranty that the library function produces always and on all platforms exactly the same numbers. Its contract seems rather to be that it produces always a sequence of numbers with apparent uniform distribution. I believe this is one of the main reasons coco has its own generator in the first place.

code-experiments/src/c_linear.c
```diff
@@ -217,9 +217,17 @@ static coco_problem_t *c_linear_shuffle(coco_problem_t *problem_c,
   
   /* Pick up a random constraint number other than 1.
    * formula for U[M,N]: 
-   * M + rand() / (RAND_MAX / (N - M + 1) + 1)
+   * random = M + (rand() / (RAND_MAX + 1.0)) * (N - M + 1)
```
nikohansen: this formula looks still wrong to me, in particular `N - M + 1` must AFAICS read `N - M`

nikohansen: BtW, what is the source of this formula?

phrsampaio: Page 161 in this [book](https://books.google.fr/books?id=7PFZAgAAQBAJ&pg=PA161&lpg=PA161#v=onepage&q&f=false) explains it.

code-experiments/src/c_linear.c
phrsampaio: No issue was created.

code-experiments/src/c_linear.c
phrsampaio: Just to make sure: this formula generates a random integer while the previous discussion was about generating a random double in another part of the code.

nikohansen: AFAICS the formula is for integers and does not produce the correct result if we aim at a continuous interval. You can easily verify this by setting `N - M == 1` that the size of the interval will be close to 2 instead of 1.

phrsampaio: That is what I said. It is exactly what is needed for this function c_linear_shuffle(). The formulae for doubles were modified in c_linear_cons_bbob_problem_allocate() and c_linear_single_cons_bbob_problem_allocate().

nikohansen: right, I didn't realize because you apparently used it for a continuous random variable (which I guess means you didn't realize as well ;-) I guess the comment should better mention this as well...

code-experiments/src/c_linear.c
nikohansen: that is a pity, it would be nice if you could create one.

phrsampaio: I will try to use coco_random_normal() again to check if it still doesn't work and create the issue if it happens so.

PR #1262, opened 2016-12-22
De-activated logloss plots and tables when no ref algorithm is provided

oaelhara:
Also merged with a more recent version of the feature-refalg branch


brockho:
[was outdated by a more recent pull request in the meantime]

Issue #1234, opened 2016-11-30
activate test_suites


nikohansen:
I factored out the regression tests of the suites in `do.py` from `test_python()` into `test_suites()`. 
What needs to be done now is to initiate this test automatically. It can be done by

```
python do.py test-suites 2 10 20
```
where the numbers are optional and refer to a data set with the respective number of solutions tested per problem, default is 2 (for a very quick test), it can be all three to test 32 solutions (for overnight test). The numbers refer to the respective test data which must be present under `test/regression-test/data`.

dtusar:
I can add this to the tests. On Jenkins I don't like having `do.py` calls with parameters since then folders with quotation marks are created. I will add a python `do.py test-suites-all` option which will test all the 32 solutions.

> The numbers refer to the respective test data which must be present under `test/regression-test/data`.

I'm not sure if I understand this correctly but if we want to run this test automatically the data must be created during the test.

nikohansen:
The data created during the test are compared against previously generated data, which are part of the repository now. I couldn't think of a better solution.

`test-suites-all` should maybe rather be called `test-suites-full`.

dtusar:
> The data created during the test are compared against previously generated data, which are part of the repository now. I couldn't think of a better solution.

This should work.

> test-suites-all should maybe rather be called test-suites-full.

OK.

dtusar:
@nikohansen On my computer the regression test for 10 solutions fails. If I run `python do.py test-suites 10` I get this error

```
PYTHON  test_suites.py 10 in code-experiments/test/regression-test
ERROR: return value=1
using file data\regression_test_10data_for_suite_bbob.py with 21600 test cases B
BOB suite problem f12 instance 58 in 2D (177, (1447.3149385050367, -830.32704880
85931)) (1.7499057709942032e+141, []) 6.09043250958e+67
Traceback (most recent call last):
  File "test_suites.py", line 83, in <module>
    "regression_test_%ddata_for_suite_" % ndata + name + ".py"))
  File "test_suites.py", line 55, in regression_test_a_suite
    assert is_equal(f(x), xfc_dict[key][0])
AssertionError
```


Can you check this before I add the `test-suites-full` option?

nikohansen:
I committed a fix for the above assert in the `development` branch. That doesn't guaranty that the remaining tests go through, so this need to be checked.

dtusar:
Similar error but for a different instance and dimension:

```
[...]

PR #1311, opened 2017-02-08
Re-gererated the large-scale regression test-data to account for changes in the function definitions (back-compatibility)

oaelhara:
Towards solving #1234


brockho:
Sorry for not merging before, but now, it seems, there is a conflict with the development branch regarding the new reference algorithm files. @oaelhara: would you be so kind and quickly do a pull from development and then open a new pull request?

oaelhara:
Done.

Issue #1343, opened 2017-04-18
bbob-constrained: implement wrappers


brockho:
Wrappers around the new C code need to be provided for the bbob-constrained suite in MATLAB, Octave, python(?), and Java. In particular, calling the constraints function should be possible.

nikohansen:
The Python wrapper should work, but if you find a problem let me know, I will take care of it.

brockho:
Can be closed now after the Java wrapper is in.

PR #1378, opened 2017-05-19
Development

asmaatamna:
Related to issue #1343 
Java wrapper now available


code-experiments/build/java/CocoJNI.c
```diff
@@ -298,16 +299,16 @@ JNIEXPORT jdoubleArray JNICALL Java_CocoJNI_cocoEvaluateConstraint
...
   x = (*jenv)->GetDoubleArrayElements(jenv, jx, NULL);
-  y = coco_allocate_vector(number_of_objectives);
+  y = coco_allocate_vector(number_of_constraints);
   coco_evaluate_constraint(problem, x, y);
 
   /* Prepare the return value */
-  jy = (*jenv)->NewDoubleArray(jenv, number_of_objectives);
-  (*jenv)->SetDoubleArrayRegion(jenv, jy, 0, number_of_objectives, y);
+  jy = (*jenv)->NewDoubleArray(jenv, number_of_constraints);
+  (*jenv)->SetDoubleArrayRegion(jenv, jy, 0, number_of_constraints, y);
```
brockho: Strange that it was like that before... Correction looks okay.

code-experiments/build/java/ExampleExperiment.java
```diff
@@ -71,7 +75,7 @@ public static void main(String[] args) {
          * Adapt to your need. Note that the experiment is run according
          * to the settings, defined in exampleExperiment(...) below.
          */
-		exampleExperiment("bbob", "bbob", randomGenerator);
```
brockho: We should keep the same functionality everywhere: hence, I would opt for going back to `exampleExperiment("bbob", "bbob", randomGenerator);`. The new `bbob-constrained` call should appear though in the comment (together with how to call the `bbob-biobj` suite probably).

code-experiments/build/java/ExampleExperiment.java
```diff
@@ -161,14 +166,16 @@ public static void exampleExperiment(String suiteName, String observerName, Rand
 	 */
 	public static void myRandomSearch(Function f, 
 			                          int dimension, 
-			                          int numberOfObjectives, 
+			                          int numberOfObjectives,
+                                      int numberOfConstraints,
```
brockho: There seems to be a wrong indentation here.

code-experiments/build/java/ExampleExperiment.java
```diff
@@ -182,6 +189,8 @@ public static void myRandomSearch(Function f,
 		    /* Call the evaluate function to evaluate x on the current problem (this is where all the COCO logging
 		     * is performed) */
 			y = f.evaluate(x);
+			if (numberOfConstraints > 0)
+				z = f.evaluateConstraint(x);
```
brockho: Usually, I would expect the algorithm to call the constraint evaluation function before the function evaluation.

code-experiments/build/matlab/exampleexperiment.m
```diff
@@ -31,7 +31,7 @@
 % bbob-biobj-ext    92 unconstrained noiseless bi-objective functions
 % bbob-largescale   24 unconstrained noiseless single-objective functions in large dimensions
 %
-suite_name = 'bbob';
+suite_name = 'bbob-constrained';
```
brockho: Same here: I think we should keep the `bbob` suite the default one and mention the `bbob-constrained` suite in the comment above instead.

code-experiments/build/matlab/my_optimizer.m
```diff
@@ -6,5 +6,20 @@ function my_optimizer (problem, lower_bounds, upper_bounds, budget)
...
+        cocoProblemGetName(problem)
     end
 end
+
+%
+% asma: The code below calls MATLAB solver, fmincon, on bbob-constrained testbed
+% cons is defined in cons.m
+%
+
+% function my_optimizer (problem, lower_bounds, upper_bounds, budget)
```
brockho: Shouldn't this go into a separate file without the comments? That allows the user to easily exchange the two solvers without uncommenting code (and without having working code within a comment).

code-experiments/build/java/CocoJNI.c
nikohansen: I am not sure to understand why would we want to remove the `number_of_objectives` call from the interface?

nikohansen: sorry, I accidentally removed the comment here. @brockho was wondering why it was like this before, while it seems OK now.

asmaatamna: Yes. This was probably a typo caused by a "copy/paste".

code-experiments/build/matlab/my_optimizer.m
asmaatamna: Now in a separate file (my_fmincon.m). I kept the comments for the moment as I find them helpful to give a quick overview of what is done by the code, but I can delete them if you prefer.

code-experiments/build/java/ExampleExperiment.java
asmaatamna: Done.

code-experiments/build/java/ExampleExperiment.java
asmaatamna: Corrected.

code-experiments/build/java/ExampleExperiment.java
asmaatamna: Done.

asmaatamna:
Tests still run through on Jenkins after the modifications. Should I close this pull request and create a new one? I don't know how to proceed...

brockho:
All the above looks okay such that we can accept the pull-request from my side.

However, one more comment:
`exampleExperiment("bbob-constrained", "bbob-constrained", randomGenerator);` does not produce any output and also does not show any warning with `python do.py run-java`. This is a bit weird since for all other test suites, the two first strings can be the same and everything works (e.g. also for the `bbob-biobj-ext` suite for which no `bbob-biobj-ext` logger exists). Seemingly, in all previous cases, an alias has been introduced for the observer name (in the C code if I remember correctly) and we should do the same for the `bbob-constrained` suite rather sooner than later.

code-experiments/build/java/ExampleExperiment.java
nikohansen: I don't think there should or need to be any prior expectation, because for many algorithms the order does not matter at all.

asmaatamna:
I did the necessary modification in coco_observer.c (the mapping is done starting from line 513). Will push the changes soon.

Issue #1393, opened 2017-05-30
evaluate_function's bookkeeping is wrong for constraints


brockho:
In the case of constraints, no restarts are triggered at the moment because the bookkeeping of the best observed `f`-value in the `evaluate_function` call does not take into account whether the evaluated point is actually feasible:
```
/* A little bit of bookkeeping */
  if (y[0] < problem->best_observed_fvalue[0]) {
    problem->best_observed_fvalue[0] = y[0];
    problem->best_observed_evaluation[0] = problem->evaluations;
  }
```

This should be fixed to get reasonable results on the `bbob-constrained` test suite.

brockho:
After the discussion with @nikohansen, @asmaatamna, @anneauger, and @dtusar, it seems as if we can simply evaluate the constraints within the above if before to update the `best_observed_fvalue` in case of a constrained problem and only do the update if the solution is feasible.

brockho:
This issue should be solved with the pull request #1406 now.

PR #1406, opened 2017-06-09
addressed #1393

brockho:


code-experiments/src/coco_problem.c
```diff
@@ -53,8 +56,18 @@ void coco_evaluate_function(coco_problem_t *problem, const double *x, double *y)
 
   /* A little bit of bookkeeping */
   if (y[0] < problem->best_observed_fvalue[0]) {
-    problem->best_observed_fvalue[0] = y[0];
-    problem->best_observed_evaluation[0] = problem->evaluations;
+    is_feasible = 1;
+    if (coco_problem_get_number_of_constraints(problem) > 0) {
+      z = coco_allocate_vector(coco_problem_get_number_of_constraints(problem));
+      /* same feasibility threshold as in logger_bbob.c did not work,
+         hence using exact feasibility: */
```
nikohansen: code and comment seem to disagree

brockho: Ups. My fault. I will remove the comment.

code-experiments/src/coco_problem.c
```diff
@@ -53,8 +56,18 @@ void coco_evaluate_function(coco_problem_t *problem, const double *x, double *y)
...
+      is_feasible = coco_is_feasible(problem, x, z, 1.0e-5);
```
nikohansen: this code should only be considered as temporary fix. The constant 1e-5 cannot stay here, as it makes a relevant difference and should not be hidden in the middle of some code.

code-experiments/src/coco_problem.c
```diff
@@ -795,7 +808,7 @@ static void coco_problem_stacked_evaluate_function(coco_problem_t *problem, cons
    */
   if (problem->number_of_constraints > 0) {
     cons_values = coco_allocate_vector(problem->number_of_constraints);
-    is_feasible = coco_is_feasible(problem, x, cons_values, 0.0);
+    is_feasible = coco_is_feasible(problem, x, cons_values, 1.0e-05);
```
nikohansen: same as above, relevant constants should/must be exposed.

code-experiments/src/suite_cons_bbob_problems.c
```diff
@@ -176,13 +176,14 @@ static coco_problem_t *f_sphere_c_linear_cons_bbob_problem_allocate(const size_t
    * objective function value into problem->best_value.
    */
   for (i = 0; i < dimension; ++i)
-    problem->best_parameter[i] = 0.0;  
-  coco_evaluate_function(problem, problem->best_parameter, problem->best_value); 
-   
-  /* Since the call to coco_evaluation_function above increments the 
-   * number of f-evaluations, reset it to zero.
+    problem->best_parameter[i] = 0.0;
+  problem->evaluate_function(problem, problem->best_parameter, problem->best_value);
```
nikohansen: consider adding an assert that problem->evaluate_function is not NULL.

code-experiments/src/coco_problem.c
```diff
@@ -29,6 +29,9 @@
 void coco_evaluate_function(coco_problem_t *problem, const double *x, double *y) {
   /* implements a safer version of problem->evaluate(problem, x, y) */
   size_t i, j;
+  size_t is_feasible;
```
nikohansen: A small comment: size_t doesn't look the right type for the is_feasible flag. coco_is_feasible returns an int, which seems perfectly fine to me. Assignment to a size_t should actually give a compiler warning that a sign may be lost.

brockho:
@nikohansen: my latest commit 7af94a5 now also addresses your latest remarks that have not yet been integrated.

PR #1457, opened 2017-07-17
Add Python 3 support to postprocessing

FloopCZ:
@ GECCO 2017 Sigevo Summer School

This adds support for Python 3 to `code-postprocessing`.
What needs to be done before merge:

- [x] Check that it still works under Python 2
- [x] Add the [six](https://pypi.python.org/pypi/six) package to requirements
- [x] Carefully review

(feel free to push in this branch)

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,9 +30,8 @@ Requirements  <a name="Requirements"></a>
   - Python >=2.6 with `setuptools` installed
   - optional: `git`
 2. For a machine running the post-processing
-  - Python 2.6 or 2.7 with `setuptools`, `numpy` (preferably >=1.7) and `matplotlib` (preferably >=2.0) installed. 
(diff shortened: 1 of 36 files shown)


nikohansen:
Thanks, very good idea to make the necessary modifications that the code at least does not bail out under Python 3! After that we would of course need to check very carefully whether it gives the same results under Python 3 than under Python 2. @islandblue may also have a look at it.

FloopCZ:
@nikohansen I have added a few more improvements and got rid of some of the warnings. Please review.

Do you have a list of requirements specified somewhere? (`requirements.txt` would be a nice to have feature, since that is a place where `pip` searches for requirements)

FloopCZ:
@nikohansen (nvm, I added `six` to the list of requirements in `README.md`.) This PR is now ready to be merged. I will make a pull request for CI in a moment.

Issue #1436, opened 2017-06-22
installing COCO within virtualenv


mashallAryan:
Hi,
How is it possible to install COCO (python) inside a virtualenv or in an alternate/custom directory 
"python do.py run-python " does not seem to provide tools for this purpose.

nikohansen:
You can use, for example, `conda` for this purpose. My expertise is limited, but AFAICS any virtual environment manager will do. After that, `python do.py` is called within this virtual environment. 

https://conda.io/docs/using/envs.html
https://uoa-eresearch.github.io/eresearch-cookbook/recipe/2014/11/20/conda/

mashallAryan:
While inside the virtualenv,  `python do.y` does not install to the virtualenv site-packages; instead tries to access the local user site-packages. Looks like   `--user` is hardcoded inside  the installation code. 

By alternative directory I mean any arbitrary directory -> in legacy python installation --target is used to do so. 
Here is the kind of error I get. I think it can be replicated easily
```
$ virtualenv env
$ source env/bin/active
$ env/bin/python do.py
...
...
...EXPAND  code-experiments/build/python/setup.py.in to code-experiments/build/python/setup.py
PYTHON  setup.py install --user in code-experiments/build/python
ERROR: return value=1
NOTE: Using precompiled C file to build interface.
running install
Checking .pth file support in ~/.local/lib/python2.7/site-packages/
~/env/bin/python -E -c pass
TEST FAILED: ~/.local/lib/python2.7/site-packages/ does NOT support .pth files
error: bad install directory or PYTHONPATH..
....
...

```

nikohansen:
Nice find. We are looking into it. As a quick fix you could try to replace line 183/188 (master/dev) of `do.py`
```python
python(join('code-experiments', 'build', 'python'), ['setup.py', 'install', '--user'])
```
with 
```python
python(join('code-experiments', 'build', 'python'), ['setup.py', 'install'])
```
and accordingly in line 241/ 246  for ``./do.py install post-processing``.

The disadvantage is that some users might not be able to install it this way, but I guess that is what `--target` could address. 

Otherwise directly calling ``code-experiments/build/python/setup.py install --target ...`` after `do.py` failed should work as well, as `do.py` has done all the necessary preparations.

mashallAryan:
after installing the coco inside virtualenv and making it relocatable (virtualenv --relocate) I get this error:
```
RuntimeError: module compiled against API version 0xb but this version of numpy is 0xa
[...]

PR #1474, opened 2017-08-01
Added option to specify Python package install location

islandblue:
We can now pass "install-user" or "install-home=_directory_" as an additional argument to change Python package install location. The default when unspecified is to install to site packages, which works with conda and should work with virtualenv or any other environment manager, I think. The former default was to install to user packages, so to get the exact same behavior as before, we need to pass "install-user" now. I've updated help with relevant usage information and marked the commands that are affected by these new arguments.

### DIFF
--- a/do.py
+++ b/do.py
@@ -236,14 +236,14 @@ def leak_check():
 
 ################################################################################
 ## Python 2
-def install_postprocessing():
+def install_postprocessing(package_install_option = []):
     ''' Installs the COCO postprocessing as python module. '''
     global RELEASE
     expand_file(join('code-postprocessing', 'setup.py.in'),
                 join('code-postprocessing', 'setup.py'),
                 {'COCO_VERSION': git_version(pep440=True)})
     # copy_tree('code-postprocessing/latex-templates', 'code-postprocessing/cocopp/latex-templates')
-    python('code-postprocessing', ['setup.py', 'install', '--user'], verbose=_verbosity)
+    python('code-postprocessing', ['setup.py', 'install'] + package_install_option, verbose=_verbosity)
 
 def test_suites(args):
     """regression test on suites via Python"""
@@ -279,21 +279,21 @@ def _prep_python():
     #     run('code-experiments/build/python/cython', ['cython', 'interface.pyx'])
 
 
-def build_python(env=None):
+def build_python(env=None, package_install_option = []):
     _prep_python()
     ## Force distutils to use Cython
     # os.environ['USE_CYTHON'] = 'true'
     # python('code-experiments/build/python', ['setup.py', 'sdist'])
     # python(join('code-experiments', 'build', 'python'), ['setup.py', 'install', '--user'])
-    python(join('code-experiments', 'build', 'python'), ['setup.py', 'install', '--user'], env=env)
+    python(join('code-experiments', 'build', 'python'), ['setup.py', 'install'] + package_install_option, env=env)
     # os.environ.pop('USE_CYTHON')
 
 
-def run_python(test=False):
+def run_python(test=False, package_install_option = []):
     """ Builds and installs the Python module `cocoex` and runs the
     `example_experiment.py` as a simple test case. If `test` is True,
     it runs, in addition, the tests in `coco_test.py`."""
-    build_python()
+    build_python(package_install_option = package_install_option)
     try:
         if test:
             python(os.path.join('code-experiments', 'build', 'python'), ['coco_test.py'])
@@ -370,8 +370,8 @@ def test_python(args=(['code-experiments/build/python', ['coco_test.py', 'None']
 
 ################################################################################
 ## Python 2
-def build_python2():
-    build_python(env='python2.7')
+def build_python2(package_install_option = []):
+    build_python(env='python2.7', package_install_option = package_install_option)
 
 
 def test_python2():
@@ -380,8 +380,8 @@ def test_python2():
 
 ################################################################################
 ## Python 3
-def build_python3():
-    build_python(env='python3')
+def build_python3(package_install_option = []):
+    build_python(env='python3', package_install_option = package_install_option)
 
 
 def test_python3():
@@ -706,12 +706,12 @@ def test_java():
 
 ################################################################################
 ## Post processing
-def test_postprocessing(all_tests=False):
-    install_postprocessing()
+def test_postprocessing(all_tests=False, package_install_option = []):
+    install_postprocessing(package_install_option = package_install_option)
     if all_tests:
         try:
             # run example experiment to have a recent data set to postprocess:
-            build_python()
+            build_python(package_install_option = package_install_option)
             python('code-experiments/build/python/', ['-c', '''
 try:
     import example_experiment as ee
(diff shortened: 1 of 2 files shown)


do.py
```diff
@@ -859,27 +859,27 @@ def help():
   build-matlab            - Build Matlab module
   build-matlab-sms        - Build SMS-EMOA example in Matlab
   build-octave            - Build Matlab module in Octave
-  build-python            - Build Python modules
-  build-python2           - Build Python 2 modules
-  build-python3           - Build Python 3 modules
-  install-postprocessing  - Install postprocessing (user-locally)
+  build-python*           - Build Python modules
```
nikohansen: The star is likely to be misunderstood by some users as part of the command which needs to be typed. The simple solution I could think of is to rather add `^*` at the end of the explanation, like
```
build-python*           - Build Python modules
```
-> 
```
build-python            - Build Python modules^*
```

islandblue: Point taken. I'm slightly uncomfortable with the fact that the explanations are of different length and some even span multiple lines, so placing the mark at the end of the explanation makes it even more inconvenient to distinguish which commands are affected by this feature at a glance, but for lack of good alternatives, I'll roll with it. Still, I suggest using `¹` instead of `^*`, just to make it look more like the familiar footnote, unless `^*` has a meaning I am unaware of.

islandblue:
We require the user of this software to be capable of so many things at lower levels than this, like coding algorithms and reading and understanding existing code. I would argue that it is somewhat inconsistent to try to hold his hand with this simpler (in my opinion) issue of user access rights, which kind of pops up many times when you install a Python package or any other software really. But it is not costly to do and I do not feel strongly about this in any direction, so here's what we can do:

- We can make `install-user` the default option, which would imitate earlier behavior, and require an additional argument to install to site packages instead. This is the most backward-compatible alternative, although I would guess a lot more people use some kind of environment manager with Python than those who work with a system-level installation, so in this case, the _default_ would be ideal only for a minority.

- We can create this scenario where the installation fails, intercept the error and inform the user of the possible cause and solutions. I do not know yet if it is possible to definitively single out the error. If not, we run the risk of interpreting anything that goes wrong with package installation as a lack of access rights and suggesting the wrong cure due to inaccurate diagnosis.

There may be other alternatives, but between these two, I'm inclined to go with the latter at first, as you too suggested, despite it being more complicated, but fall back to the former if it is not feasible to accurately identify the exact fault.

And there's the fact that informing the user about the problem and the possible solutions will still cause our test cases to break, so we would need to modify our tests to run with one of the additional arguments anyway.

do.py
nikohansen: Agreed `^1` is probably a better choice.

nikohansen:
> We require the user of this software to be capable of so many things at lower levels than this, like coding algorithms and reading and understanding existing code. I would argue that it is somewhat inconsistent to try to hold his hand with this simpler (in my opinion) issue of user access rights, which kind of pops up many times when you install a Python package or any other software really. 

That is not my understanding of the COCO platform and I think it is important to be on the same page on this one. A user may

- never have touched, seen or written a single line of Python (but may be a heavy Matlab user, or well versed in C, or neither). 
- only want to have a look at COCO data without any specific programming experience or knowledge

Biologists or chemists or engineers or mathematicians are definitely in the group of potential users who could have any kind of software skills or lack thereof. Even if they could acquire the necessary skills, it would be from their perspective an unnecessary time sink in the best case, or they just give up in the worst case (from our perspective).

nikohansen:
> If not, we run the risk of interpreting anything that goes wrong with package installation as a lack of access rights and suggesting the wrong cure due to inaccurate diagnosis.

On a general note, it is IMHO quite rarely a good idea to dismiss an approximate solution because one cannot figure out the optimal solution. 

Specifically, I can hardly believe that anybody would expect a software bailing out and suggesting a fix gets the suggestion right 100% of the time, even if it would claim to be right 100% of the time (which it shouldn't unless for very good reasons).

Even a software as technical as [MacPorts](https://www.macports.org/) pleases us with a message like

> Error: Insufficient privileges to write to MacPorts install prefix.

instead of just bailing out on the omission of `sudo`. In our case, IMHO a more verbose message which includes a possible solution (i.e. an answer to the question "what do I do now/next?") is in order.

brockho:
I totally agree with Niko's last two comments. We should make/keep COCO as user-friendly as possible and as users, we should have in mind all potential users, in particular the ones from other scientific fields as Niko indicated.

islandblue:
You will have zero argument and zealous support from me in trying to make/keep COCO user-friendly. I was actually under the impression that you'd drawn the red line in user-friendliness closer than I would have personally preferred. It is utterly meaningless to defend my perceived view of your standpoint against your own direct statement, so I merely stand corrected, and gladly so, but for the sake of being on the same page, I propose that, during the next meeting, we discuss once more the couple observations that led me astray in my perception of your opinion on this matter.

That being said, should I take your argument on user-friendliness as an indication that you are in favor of keeping site packages as the default install location, instead of keeping the old default (user)?

nikohansen:
> That being said, should I take your argument on user-friendliness as an indication that you are in favor of keeping site packages as the default install location, instead of keeping the old default (user)?

See my comments https://github.com/numbbo/coco/issues/1436#issuecomment-313547238 and https://github.com/numbbo/coco/issues/1436#issuecomment-313633529. Inheriting the default behaviour from `pip` seems to me a very reasonable choice.

nikohansen:
> I propose that, during the next meeting, we discuss once more the couple observations that led me astray in my perception of your opinion on this matter.

Of course! I am sorry that I wasn't as clear as I probably should have been.

islandblue:
I conjured up access issues by playing with permissions of the site packages directory under my conda environment. Wrapping the call to `setup()` in `setup.py.in` seemed like a good idea but the output there gets swallowed somehow, so I switched to an alternative and augmented the `python()` function in `cocoutils.py` to allow for an optional custom exception handler. I wrote up a handler in `do.py` and passed it to all calls to `python()` that invoke `setup.py` with the `install` command.
The `¹` character I suggested using gives birth to encoding issues and is thus more trouble than it's worth. Instead of shuffling through cryptic symbols, I simply added `(see NOTE below)` to the descriptions of all commands involved. Also trimmed part of the help text to 79 characters for good measure.

nikohansen:
I have one question regarding
```python
        elif arg == 'install-user':
             package_install_option = ['--user']
```
Why is the option `install-user` and not the original option string `--user` as `pip` needs to see it? Thinking further, even any option could be passed to `pip` then.

islandblue:
Very deliberately. I chose the safer (in my opinion) approach by granting the user just as much freedom as was absolutely necessary, to protect the user from himself. If we allow anything to be passed to pip, we would have to test with every possible pip argument to make sure the user cannot shoot himself in the leg. It is not more difficult to just use pip arguments and pass them directly, but it could lead to complications. Another point was that we also pass arguments to do.py (i.e. `and-test`) and might want to pass even more in the future, so we need to distinguish between arguments for do.py and arguments to be passed to pip. Passing everything with a dash to pip is one way to go. Just let me know if you think it is better and I'll fix it immediately.

nikohansen:
> we would have to test with every possible pip argument to make sure the user cannot shoot himself in the leg

As `pip` is a quite generic tool for non-administrators, I don't think this would be necessary. 

Passing everything starting with a double dash seems a reasonable possibility (all options in `pip` have a double dash version). Seems like a low priority issue though.

islandblue:
Yes, but easy to do and since I'm already adding the sudo bit to the help text, I'll get it out of the way while I'm at it.

nikohansen:
@islandblue It looks like some of the stars in `do.py`, like `build-python*` (line 895) are still there in the development branch. Not sure whether this happened in the pull.

nikohansen:
Re `pip`: I got this wrong, `do.py` does not call `pip` put `setup.py` which calls `distutils.core.setup`. Then I suggest we leave the options as they are for the time being.

PR #1478, opened 2017-08-06
[WIP] Improve continuous integration

FloopCZ:
Hi @nikohansen and @brockho. Now that you have a new CI set up, we should make it useful. I suggest to run all the available COCO unit tests. I have already tried to add your unit tests to `.circleci/config.yml` and `appveyor.yml`, however, I struggle to make the tests pass.

1. Windows fails to pass `test-c` with [this error](https://ci.appveyor.com/project/FloopCZ/coco/build/1.0.54/job/4eet7bdrtd4d5r9u) (WindowsError: [Error 2] The system cannot find the file specified). I am not sure what file is it missing, maybe there is something more that has to be done before running `test-c`? @brockho will probably know more.
2. Ubuntu Python 2 fails to pass postprocessing tests with [this error](https://circleci.com/gh/FloopCZ/coco/176) (Test failed: error while running bibtex on templateBBOBarticle.). I may be missing some special latex dependency installed, but I have no Idea which one.
3. Ubuntu Python 3 fails to pass some preprocessing or postprocessing tests with [this error](https://circleci.com/gh/FloopCZ/coco/177) (ImportError: No module named setuptools). This one may be caused by the fact that somewhere in the preprocessing/postprocessing code, python 2 is hardcoded (instead of using the python interpreter, which is used to run the main script). I suppose that this error does not appear in Jenkins and your personal computers, since they have both Python 2 and 3 installed. However, in the CircleCI environment, I have deliberately installed only the Python that is really tested.

If you are interested, please push some more edits to `.circleci/config.yml` and `appveyor.yml` into this branch to make it pass all the tests. You have the permission to push into it. I believe that you are more experienced with your own unit tests.

P.S.: I had to put in a few more Python 2 to Python 3 conversions to make the tests compile under Python 3, but nothing serious.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,31 +2,31 @@ version: 2
 
 references:
 
-  test_ubuntu_python3: &test_ubuntu_python3
+  install_ubuntu_python2: &install_ubuntu_python2
     run:
-      name: Run tests on Ubuntu with Python 3.
+      name: Install Python 2 on Ubuntu.
       command: |
-        # install dependencies
         apt-get update -qy
-        apt-get install -y build-essential python3-dev python3-numpy python3-matplotlib \
-                           python3-six python3-setuptools python3-scipy
-        # install coco
-        python3 do.py run-python
-        python3 do.py install-postprocessing
-        # run the example
-        cd code-experiments/build/python
-        python3 example_experiment.py bbob
-        # postprocess the results
-        python3 -m cocopp -o ./postproc ./exdata/*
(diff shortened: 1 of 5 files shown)


nikohansen:
@dtusar , I suggest to merge and go from there? It looks to me like you have a pretty good grasp on how to fix any of the above failures, am I wrong?

dtusar:
I agree since the failures are not code related.

nikohansen:
Can we convert the above (first) comment into an issue?

Issue #1350, opened 2017-04-19
Warnings during build-python


brockho:
We still have some warnings that show up when building the python wrapper of COCO. I guess that the same appear when compiling C with a strict enough compiler:

```
code-experiments/src/suite_bbob_legacy_code.c(84) : warning C4018: '<' : signed/unsigned mismatch
code-experiments/src/suite_bbob_legacy_code.c(145) : warning C4018: '<' : signed/unsigned mismatch
code-experiments/src/suite_bbob_legacy_code.c(148) : warning C4018: '<' : signed/unsigned mismatch
code-experiments/src/suite_bbob_legacy_code.c(150) : warning C4018: '<' : signed/unsigned mismatch
code-experiments/src/suite_bbob_legacy_code.c(154) : warning C4018: '<' : signed/unsigned mismatch
code-experiments/src/suite_bbob_legacy_code.c(156) : warning C4018: '<' : signed/unsigned mismatch
code-experiments/src/suite_bbob_legacy_code.c(180) : warning C4018: '<' : signed/unsigned mismatch
code-experiments/src/large_scale_transformations.c(181) : warning C4018: '<' : signed/unsigned misma
tch
code-experiments/src/large_scale_transformations.c(215) : warning C4018: '<' : signed/unsigned misma
tch
code-experiments/src/large_scale_transformations.c(225) : warning C4018: '<' : signed/unsigned misma
tch
code-experiments/src/large_scale_transformations.c(291) : warning C4018: '<' : signed/unsigned misma
tch
code-experiments/src/transform_vars_permblockdiag.c(84) : warning C4018: '<' : signed/unsigned misma
tch
code-experiments/src/transform_vars_permblockdiag.c(99) : warning C4018: '<' : signed/unsigned misma
tch
code-experiments/src/transform_vars_permblockdiag.c(100) : warning C4018: '>=' : signed/unsigned mis
match
code-experiments/src/f_sharp_ridge.c(32) : warning C4244: '=' : conversion from 'double' to 'size_t'
, possible loss of data
code-experiments/src/c_linear.c(229) : warning C4244: '=' : conversion from 'double' to 'size_t', po
ssible loss of data
code-experiments/src/suite_cons_bbob.c(64) : warning C4090: 'function' : different 'const' qualifier
s
code-experiments/src/suite_cons_bbob.c(130) : warning C4090: 'function' : different 'const' qualifie
rs
[...]

PR #1506, opened 2017-08-28
address last warning of issue #1350

brockho:
Here is a suggestion to take care of the `const double *x` warning in `logger_bbob_evaluate` of issue #1350. @asmaatamna and @nikohansen: because it is an important change, it will be nice if you can have a look at the code before we merge it with the development branch.


code-experiments/src/logger_bbob.c
```diff
@@ -447,31 +453,53 @@ static void logger_bbob_evaluate(coco_problem_t *problem, double *x, double *y)
...
+    }
+    z_objvalue = y[0];
+  }
+  else {
+	  /* No logging if not feasible or not better than best. */
+		coco_free_memory(z);
+		return;
+  }
+
+	printf("z_objvalue = %e\n", z_objvalue);
```
nikohansen: Are you sure these `print`s were meant to stay?

brockho: Ups. Sorry. Many thanks for catching this. The current version shouldn't have these lines anymore.

Issue #35, opened 2015-08-08
Allow for .zip files to contain data for postprocessing


GoogleCodeExporter:
```
(after new data format is introduced) processInputArgs/findfiles should also 
accept zipped archives (now it accepts 'tar', '.tar.gz'). Maybe: write a small 
class, that opens a file or folder regardless wether it is zipped, tarred, 
tar.gz... 
```

Original issue reported on code.google.com by `dimo.bro...@inria.fr` on 13 Jun 2014 at 5:55

brockho:
replaces #9

brockho:
Seems to work in the development branch for about two weeks now without complaints. I hence close the issue.

PR #1509, opened 2017-08-28
closes issue #35

brockho:
Unfortunately, I still see many error leaks via valgrind on the Mac CI slave (devel-test2 branch). Don't know whether this is due to the bad support for OSX or really my fault (did not change anything on the `C` side).


brockho:
The leaks are gone in the latest tests (no change in the new code, just pulled from `development`). The CircleCI and Appveyor tests look like they show the same errors as for the current `development` branch. Hence, we should probably merge this pull request as well asap.

The only Jenkins test that failed looks like a Jenkins failure. I restarted the test a few minutes ago to see whether this is actually the case or whether it is caused by some changes in the code.

brockho:
Update: all Jenkins tests now run through (it was actually a Jenkins problem before). I suggest merging in the coming days.

nikohansen:
I would really like to see the Python 2 test on circleci go back to green.

PR #1507, opened 2017-08-28
Reset foreground algorithm list if cocopp.main() is called.

islandblue:
The issue itself is resolved. Still prints out that foreground algorithm list is changed from empty set to whatever, though. See second last line in the output below. Should read `foreground_algorithm_list: from ['.\\CMAES', '.\\DEAE'] to ['.\\CMAES', '.\\JADE']`, in my opinion. If that's any concern, we can open an issue for it, I think.

```
In [2]: import cocopp

In [3]: cocopp.main("CMAES DEAE")

Post-processing (2+): will generate output data in folder ppdata
  this might take several minutes.
  Data consistent according to consistency_check() in pproc.DataSet
  using: .\CMAES
  Data consistent according to consistency_check() in pproc.DataSet
  using: .\DEAE
ECDF runlength ratio graphs...
C:\Users\randopt\AppData\Roaming\Python\Python27\site-packages\cocopp-2.0.528-py2.7.egg\cocopp\comp2\pprldistr2.py:146: RuntimeWarning: invalid value encountered in divide
  x.append((tmp1/tmp0).flatten())  # inf/inf results in nan
  done (Mon Aug 28 16:32:38 2017).
ECDF runlength graphs...
  done (Mon Aug 28 16:32:51 2017).
ECDF graphs per noise group...
Loading best algorithm data from refalgs/best2009-bbob.tar.gz ...
  Data consistent according to consistency_check() in pproc.DataSet
  using: C:\Users\randopt\AppData\Roaming\Python\Python27\site-packages\cocopp-2.0.528-py2.7.egg\cocopp\refalgs/best2009-bbob.tar.gz
  done (Mon Aug 28 16:32:52 2017).
  done (Mon Aug 28 16:32:55 2017).
ECDF graphs per function group...
  done (Mon Aug 28 16:33:06 2017).
ECDF graphs per function...
  done (Mon Aug 28 16:33:50 2017).
Generating comparison tables...
  done (Mon Aug 28 16:33:52 2017).
Scatter plots...
  done (Mon Aug 28 16:34:09 2017).
Scaling figures...
  done (Mon Aug 28 16:34:24 2017).
Changed settings in `genericsettings` (compared to default):
    foreground_algorithm_list: from [] to ['.\\CMAES', '.\\DEAE']
ALL done (Mon Aug 28 16:34:24 2017).

In [4]: cocopp.main("CMAES JADE")

Post-processing (2+): will generate output data in folder ppdata
  this might take several minutes.
[...]

PR #1511, opened 2017-08-29
Updated setting if the scatter plots will be generated.

dtusar:


dtusar:
> 
>     If the changed removed a bug, it looks like it left two more bugs unaddressed (next two lines).

I agree, all three should be done the same way.
 
>     the variable genericsettings.isScatter now (still) exists but does not reflect what is actually happening, in the best case it is never touched anywhere.

In my opinion the values in `genericsettings` apply for all runs and if you change them the change stays until you change them back. But the command line flag should affect only the current run.

brockho:
I would also rather argue like @dtusar in his last comment: `genericsettings` should apply to everything, direct changes should stay until changed (back) and genericsettings might be overwritten temporarily for the current run by command line options like `--tab-only`.

Maybe, we can have a `genericsettings.py.in` in the repository that is copied to `genericsetting.py` on installation and the only purpose of which would be to make the default settings easier to retrieve while working interactively with COCO? [maybe there is also a better way of doing so]

brockho:
Thinking again about it, I would like to change my opinion :-) The above actually does not hold for `foreground_algorithm_list`, see also #1507. The reason is that the foreground algorithms are the mandatory input to `cocopp.main`.

dtusar:
Maybe it could hold also for the foreground algorithms and then they wouldn't be mandatory if specified in `genericsettings`.

nikohansen:
> Maybe, we can have a genericsettings.py.in in the repository that is copied to genericsetting.py on installation and the only purpose of which would be to make the default settings easier to retrieve while working interactively with COCO? [maybe there is also a better way of doing so]

You mean like [this](https://github.com/numbbo/coco/blob/development/code-postprocessing/cocopp/rungeneric.py#L188)?

brockho:
Yes, didn't know that we have this already. Would be nice to have this `stored_settings` available as `default_settings` at some point (low priority I would guess because we can also look into the `genericsettings.py` file ourselves).

nikohansen:
Good point, looks like a simple fix of the above hack in `rungeneric.py`.

Issue #1520, opened 2017-08-30
CI broken


nikohansen:
This is what I have found so far: 

OK: Aug 29, 2017 10:13:48 PM 4b1497a0e6d4de4a0dd75e03779d6c5349fa21ae 
BROKE:  Aug 29, 2017 10:28:26 PM 530d79d84c00ac237750a253d1859d273e55194f 
```
git show 530d79d84c00ac237750a253d1859d273e55194f

commit 530d79d84c00ac237750a253d1859d273e55194f
Merge: e1962930 32d8feba
Author: brockhof <dimo.brockhoff@inria.fr>
Date:   Tue Aug 29 22:14:58 2017 +0200

    Merge branch 'development' of https://github.com/numbbo/coco into dimo-issue35
```
But I can't say for sure that this reflects our current `development` branch.

brockho:
Yeah. That's a difficulty with CircleCI: you never know from which branch the error comes.

However, I have in mind that this problem came from running some new code on the test branches. The code of the test is most likely not in the development branch already because the `dimo-issue35` branch addresses, as the name suggests, issue #35 and is not yet merged with development (open pull request #1509).

I will run the test on this `dimo-issue35` branch once more on a test branch right now to be sure.

brockho:
The dimo-issue35 branch now breaks more tests than before (I pulled before from the current development branch though).

What is very strange is that the doctests fail for example here:
```
File "E:\builds\workspace\Coco-test1-win\dotype\test-postprocessing\node\numbbo-win\code-postprocessing\cocopp\pproc.py", line 721, in pproc.py
Failed example:
    dslist2 = bb.load(infoFile)
Expected:
      Data consistent according to consistency_check() in pproc.DataSet
Got:
      Data consistent according to test in consistency_check() in pproc.DataSet
```
although when I run the same in an IPython shell myself (on my laptop with a fresh `do.py install-postprocessing` from this branch), I get the expected output.

nikohansen:
Yes, weird, I suspect that there was an unintentional merge which introduced lots of changes. Maybe it is the problem we expected to see at some point (rarely, but even not never) when merging from (and in particular into) the devel-test branches.

This is the diff from the latest branch which was tested OK an hour ago and the development branch in my computer c94b0a0d: 
```
git diff 4b1497a0e6d4de4a0dd75e03779d6c5349fa21ae --stat
 .circleci/config.yml                                                                  |    99 +
 AUTHORS                                                                               |     1 +
[...]

PR #1521, opened 2017-08-30
test dimo-issue35 branch once more (see #1520)

brockho:
I suspect that we will see the same valgrind errors as in the open pull request #1509.


brockho:
Interestingly the valgrind error is gone now. What we see seems to be the same error in `test-postprocessing` than in the development branch.

Issue #863, opened 2016-03-21
Suite options with extra spaces don't produce a warning


ttusar:
If, for example, suite options are written like this:
`"dimensions: 2, 3, 5, 10 function_indices: 20"`
The experiments will be run only for dimension 2 without any warning.

Issue first mentioned in #724

nikohansen:
I would consider this as a bug as long as not even a warning is produced.

brockho:
Seems to work most of the times, indeed.

The above example also works for me as does `"dimensions: 2, 3, 5, 10 function_indices: 20, 21"`:
```
COCO INFO: Results will be output to folder exdata\random_search_on_bbob_budget0001xD-007
Benchmarking solver '<function random_search' with budget=1*dimension on bbob suite, Thu Aug 31 17:4
6:57 2017
Aug31 17h46:57, d=2, running: f020...............(15)  f021...............(15)
    done in 6.0e-03 seconds/evaluation

Aug31 17h46:58, d=3, running: f020...............(15)  f021...............(15)
    done in 3.9e-03 seconds/evaluation

Aug31 17h46:58, d=5, running: f020...............(15)  f021...............(15)
    done in 2.8e-03 seconds/evaluation

Aug31 17h46:59, d=10, running: f020...............(15)  f021...............(15)
    done in 1.4e-03 seconds/evaluation
  dimension seconds/evaluations
  -----------------------------
      2      6.0e-03
      3      3.9e-03
      5      2.8e-03
     10      1.4e-03
  -----------------------------
  bbob done (120 of 120 problems benchmarked), Thu Aug 31 17:46:59 2017 (0h00:01 total elapsed time)
```

However if I run it with `"dimensions: 2, 3, 5, 10 function_indices: 20"` the output is not showing the `f20` in all but dimension 2 and summarizes the total number of instances run wrongly:
```
COCO INFO: Results will be output to folder exdata\random_search_on_bbob_budget0001xD-005
Benchmarking solver '<function random_search' with budget=1*dimension on bbob suite, Thu Aug 31 17:4
5:52 2017
Aug31 17h45:52, d=2, running: f020...............
    done in 5.3e-03 seconds/evaluation

Aug31 17h45:53, d=3, running: ...............
    done in 3.9e-03 seconds/evaluation

Aug31 17h45:53, d=5, running: ...............
    done in 2.6e-03 seconds/evaluation

Aug31 17h45:53, d=10, running: ...............(60)
[...]

PR #1522, opened 2017-08-31
closes issue #863

islandblue:
From what I could garner looking at the usage of this function, names always start with alphabetic characters and values consist of non-alphabetic characters. With a couple extra assumptions in this direction, which conform to the present usage, the parser is now immune to white-space.

One unseemly behavior could be that `something: 1, 2, 3 4, 5` is read as `[1, 2, 34, 5]`, but that's bad format anyway. Let me know if we'd like that handled.


ttusar:
The changes look good to me.

nikohansen:
Looks like an improvement to me and solves the issue. I think @ttusar should have a quick look as well. 

> but that's bad format anyway.

the objective would still be to detect as many user errors as possible with reasonable effort, even if they fall under bad format.

A general remark on the functions `coco_options_read_*(const char *options, const char *name, char *pointer)`: it looks scary to have no information about the allocated space writable under `pointer`.

nikohansen:
Can you comment on what happens to the following input string? 
```
"dimensions: 2, 3, 5, 10 function_indices: 20, 21"
```

islandblue:
> Can you comment on what happens to the following input string?

@nikohansen, could you elaborate?

ttusar:
> I think @ttusar should have a quick look as well.

I did have a look at the code, should I do something else?

nikohansen:
> could you elaborate?

I am trying to understand why in `"dimensions: 2, 3, 5, 10 function_indices: 20, 21"` the space between `10` and `funct...` is not eaten away, or maybe it is but doesn't pose a problem afterwards?

islandblue:
The earlier code interpreted white-space as the end of the value string. The proposed code simply drops all white-space and keeps reading in the rest until it encounters an alphabetic character or the null terminator. Therefore values and following names will never fuse, because values do not have letters in them.

islandblue:
> the objective would still be to detect as many user errors as possible with reasonable effort, even if they fall under bad format.

That is indeed the objective, but to be able to identify this error, additional assumptions are required.

Presently, all this function has to know is...
- names start with an alphabetic character, which can be followed by any sequence of characters
- names are followed by a colon
- values consist of non-alphabetic characters
...however, to be able to identify errors, it needs more information, such as...
- which characters serve as delimiters between elements in an array value
- what other non-delimiter characters are allowed and their expected syntax
...which will expand the definition (and therefore the complexity) of the function significantly.

I'm not saying it will be a complex function in an absolute sense. It will still be a very elementary parsing function. But it will not be what it was before. Please confirm this is what we want, I'll be happy to do it.

brockho:
This issue can be closed after discussing it again today. The current status is enough and we do not want to add complexity to the code to check for other wrong usages.

islandblue:
Looks to me like this pull request was closed without merging to development. I believe this to be in error. @brockho can you confirm?

brockho:
@islandblue: right, the pull request was closed before merging. I apologize for this and will merge it now.

PR #1527, opened 2017-09-04
Removed all explicit python environment settings from do.py.

dtusar:


do.py
```diff
@@ -972,6 +948,9 @@ def main(args):
             package_install_option = ['--user']
         elif arg[:13] == 'install-home=':
             package_install_option = ['--home=' + arg[13:]]
+    # as long as our tests brake otherwise, '--user' remains default:
```
islandblue: @dtusar These lines were removed from development. Is this intentional?

dtusar:
No, my mistake. Thank you for noticing.

Issue #1523, opened 2017-08-31
python -m cocopp.test fails (breaking circleci)


nikohansen:
The python 2 test on circleci [fails](https://circleci.com/gh/numbbo/coco/298) since a few days (it was OK before). I get the same error on my Mac when invoking `python -m cocopp.test` from the `code-postprocessing` folder. The test bails out when doctest is run. 

```
[...]
ALL done (Thu Aug 31 14:14:44 2017).
**  subtest 2 finished in  164.675256968  seconds
launching doctest (it might be necessary to close a few pop up windows to finish)
cocopp/pproc.py:2728: UserWarning: No data for algorithm data/MCS.tgz in 40-D.
  warnings.warn(txt)
cocopp/bestalg.py:133: UserWarning: Algorithm data/MCS.tgz was not tested on f1 40-D.
  % (alg, f, d))
cocopp/bestalg.py:133: UserWarning: Algorithm data/MCS.tgz was not tested on f2 40-D.
  % (alg, f, d))
cocopp/bestalg.py:133: UserWarning: Algorithm data/MCS.tgz was not tested on f3 40-D.
  % (alg, f, d))
cocopp/bestalg.py:133: UserWarning: Algorithm data/MCS.tgz was not tested on f4 40-D.
  % (alg, f, d))
cocopp/bestalg.py:133: UserWarning: Algorithm data/MCS.tgz was not tested on f5 40-D.
  % (alg, f, d))
cocopp/bestalg.py:133: UserWarning: Algorithm data/MCS.tgz was not tested on f6 40-D.
  % (alg, f, d))
cocopp/bestalg.py:133: UserWarning: Algorithm data/MCS.tgz was not tested on f7 40-D.
  % (alg, f, d))
cocopp/bestalg.py:133: UserWarning: Algorithm data/MCS.tgz was not tested on f8 40-D.
  % (alg, f, d))
cocopp/bestalg.py:133: UserWarning: Algorithm data/MCS.tgz was not tested on f9 40-D.
  % (alg, f, d))
cocopp/bestalg.py:133: UserWarning: Algorithm data/MCS.tgz was not tested on f10 40-D.
  % (alg, f, d))
cocopp/bestalg.py:133: UserWarning: Algorithm data/MCS.tgz was not tested on f11 40-D.
  % (alg, f, d))
cocopp/bestalg.py:133: UserWarning: Algorithm data/MCS.tgz was not tested on f12 40-D.
  % (alg, f, d))
[...]

PR #1528, opened 2017-09-05
solves #1523

brockho:
In order to do so, I
   * added `from . import config` in `__init__.py`
   * added `relative_load=False` when reference algorithm is loaded in `pproc.py`
In addition, I cleaned up paths in doctests wherever `os.chdir` is used (needs to be made nicer, i.e. by removing usage of `os.chdir` as much as possible).

### DIFF
--- a/code-postprocessing/cocopp/pproc.py
+++ b/code-postprocessing/cocopp/pproc.py
@@ -300,7 +300,7 @@ class RunlengthBasedTargetValues(TargetValues):
             self._short_info = 'reference budgets from ' + self.reference_algorithm
 
             from . import bestalg
-            self.reference_data = bestalg.load_reference_algorithm(self.reference_algorithm, force=True)
+            self.reference_data = bestalg.load_reference_algorithm(self.reference_algorithm, force=True, relative_load=False)
             # TODO: remove targets smaller than 1e-8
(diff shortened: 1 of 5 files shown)


nikohansen:
I don't quite understand why changing the `relative_load` flag cannot break other use cases which seemed to work so far. Any suggestions?

code-postprocessing/cocopp/pproc.py
```diff
@@ -300,7 +300,7 @@ def initialize(self):
             self._short_info = 'reference budgets from ' + self.reference_algorithm
 
             from . import bestalg
-            self.reference_data = bestalg.load_reference_algorithm(self.reference_algorithm, force=True)
+            self.reference_data = bestalg.load_reference_algorithm(self.reference_algorithm, force=True, relative_load=False)
```
nikohansen: This change breaks the code when executed like
```
cocopp.main('--runlength-based ...')
```
from a Python shell.

nikohansen:
It turns out that the above mentioned change (L303 in `pproc.py`) does break the code when calling
```python
cocopp.main('--runlength-based ...'
```
This raises two issues to be fixed
1. test coverage of this case, that is, the test should currently not run through OK
1. after that, the code must be fixed that the test runs fine

Issue #1497, opened 2017-08-25
post-processing output folder message


nikohansen:
We see the message 
```
Post-processing (2+): will generate output data in folder ppdata
```
however this isn't too helpful when the `ppdata` folder looks for example like this:
<img width="430" alt="screen shot 2017-08-25 at 17 47 14" src="https://user-images.githubusercontent.com/7316439/29721461-949875fa-89bd-11e7-9188-b79d91f8b786.png">

Hence, information on the written subfolder should be printed as well!

nikohansen:
FTR, this now happens much less often after some changes (not sure what the change was), but it still happens once in awhile (not sure what the precise conditions are). Any hints from anybody?

islandblue:
Confused. I was under the impression that this is not something that happens but just the way output is. Did I get the issue wrong? Don't we just want to add the subfolder to the informative text?

nikohansen:
Sorry, I was too fast, the above comment refers to https://github.com/numbbo/coco/issues/1472

nikohansen:
I still don't see which folder I have to open:
```
import glob
import cocopp
datahome = '/Users/hansen/svn/coco/BBOB/data-archive/data/gecco-bbob-1-24/'
datafiles = glob.glob(datahome + '2009/rawdata/*M*tgz')
dsl = cocopp.main('--runlength-based ' + ' '.join([df for df in datafiles]))

[...]

Loading best algorithm data from refalgs/best2009-bbob.tar.gz ...
    archive extracted to folder cocopp/refalgs/.extracted_best2009-bbob ...
  Data consistent according to consistency_check() in pproc.DataSet
  using: cocopp/refalgs/best2009-bbob.tar.gz
  done (Wed Sep  6 10:19:56 2017).
  done (Wed Sep  6 10:20:19 2017).
Setting changes in `cocopp.genericsettings` compared to default:
    runlength_based_targets: from False to True
    foreground_algorithm_list: from [] to ['/Users/hansen/svn/co...
ALL done (Wed Sep  6 10:20:20 2017).
```

islandblue:
Will add after resolving a conflict in the development branch.

PR #1529, opened 2017-09-06
Print output folder at the end of post-processing (issue #1497)

islandblue:
Should print now for both single and multiple algorithms.

### DIFF
--- a/code-postprocessing/cocopp/rungenericmany.py
+++ b/code-postprocessing/cocopp/rungenericmany.py
@@ -596,6 +596,8 @@ def main(argv=None):
                         latex_commands_file)
             plt.rcdefaults()
             print_done()
+        if prepare_figures or prepare_tables or prepare_RLDistr or prepare_scatter:
+            print("Output data written to folder %s" % many_algorithms_output)
 
         plt.rcdefaults()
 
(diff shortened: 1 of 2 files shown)


nikohansen:
We have currently this intricate critical bug of #1523 we couldn't solve for several days, so I hesitate to accept any pull request before this is solved.

code-postprocessing/cocopp/rungenericmany.py
```diff
@@ -596,6 +596,8 @@ def main(argv=None):
                         latex_commands_file)
             plt.rcdefaults()
             print_done()
+        if prepare_figures or prepare_tables or prepare_RLDistr or prepare_scatter:
```
nikohansen: Why the condition? It seems to be inconsistent with line 292, where this output is printed unconditionally. Isn't it also bug prone to condition the output on a list which could easily extend when the code evolves and hence break this line?

brockho: I also would think that the output should always appear.

islandblue: I agree completely.

A similar line was already there in rungeneric1.py, so I chose to follow suit in rungenericmany.py. There may have been reasons to make it conditional like this. It would have taken time to investigate. For now, I prefer to address the actual subject of the issue. I try to avoid turning the issue into an excuse to question why the code is there the way it is. There will be a rewrite anyway. All code will go through review before making it to the new code base. Made sense to me to minimize effort spent on code that will likely not be preserved.

That being said, since we already did go the distance for this specific bit and we all agree on what is better, I will remove the condition.

Issue #1518, opened 2017-08-29
Logging the initial solution


nikohansen:
It seems that the logger records the initial solution, if the first evaluation is not better: 

https://github.com/numbbo/coco/blob/development/code-experiments/src/logger_bbob.c#L440

IIRC, this has not been the case before the constrained suite was introduced. I understand that this is a reasonable workaround if the first evaluation is infeasible. I don't see however why we should do this unconditionally (in particular on unconstrained suites but also on constrained suites). It only removes information we would otherwise have, while I don't see an advantage to it.

Any opinions?

brockho:
If I understand (my own :-) ) code correctly, it does not change anything in the unconstrained case (i.e. no replacement of the first evaluation) because of the condition `if (logger->number_of_evaluations == 1 && problem->number_of_constraints > 0)` in line 415.

The reason for this replacement is the fact that the postprocessing expects an entry in the `.dat` files with the evaluation counter `1`, IIRC. Because we do not want to write infeasible solutions, hence, the replace the first evaluated point with the feasible provided solution if it itself is not feasible.

nikohansen:
Ouups, I missed the condition in line 415, so the unconstrained case seems fine. 

Have you checked line 423 and following? Do they play well with the rest of the code? Currently I don't see why this is actually necessary (I also don't see that it is harmful).

brockho:
Lines 423 and following are as well mainly to ensure that a line with evaluation counter 1 is written in the constrained case (evaluation counter = sum of f- and g-evaluations).

I am not so sure, though, how this goes with the new data format of #1367. Maybe it won't be necessary anymore, but it might be still a good idea: at the moment, we want to always store the first `f`- as well as the first `g`-evaluation in order to allow later to display just only `g` or just only `f` evaluations (or like now the sum of both) with the guarantee of an entry for the time "1". In particular, we want to cover the case where the first `g`-evaluation happens before the first `f`-evaluation. That's why we have lines 423ff.

nikohansen:
> In particular, we want to cover the case where the first g-evaluation happens before the first f-evaluation. That's why we have lines 423ff.

There again is line 415: lines 423ff only applies when `evaluations==1`, not when `evaluations==0`. 

[...]

PR #1537, opened 2017-09-07
Development

asmaatamna:
Testing changes to logger_bbob.c

### DIFF
--- a/code-experiments/src/logger_bbob.c
+++ b/code-experiments/src/logger_bbob.c
@@ -121,7 +121,7 @@ static void logger_bbob_write_data(FILE *target_file,
   if (constrained_problem) {
     fprintf(target_file, "%lu %lu %+10.9e %+10.9e %+10.9e %+10.9e", (unsigned long) number_of_f_evaluations,
     		(unsigned long) number_of_cons_evaluations, fvalue - best_value, best_fvalue - best_value,
-            fvalue, best_fvalue);
+    		fvalue, best_fvalue);
   } else {
     fprintf(target_file, "%lu %+10.9e %+10.9e %+10.9e %+10.9e", (unsigned long) number_of_f_evaluations,
     		fvalue - best_value, best_fvalue - best_value, fvalue, best_fvalue);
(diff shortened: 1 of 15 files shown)


code-experiments/src/logger_bbob.c
```diff
@@ -426,12 +443,14 @@ static void logger_bbob_evaluate(coco_problem_t *problem, const double *x, doubl
        * does not correspond to evaluation counter 1.
        */
       logger_bbob_write_data(logger->fdata_file, 
-          1, initial_solution_fvalue, initial_solution_fvalue, 
-          logger->optimal_fvalue, inner_problem->initial_solution, problem->number_of_variables);
+          1, 0, initial_solution_fvalue, initial_solution_fvalue,
```
nikohansen: I don't see why `0` is correct here. Shouldn't it be `logger->number_of_evaluations_constraints`?

Is there a problem with the first entry having non-zero or even a large constraints evaluation counter? I wouldn't be entirely comfortable with ignoring them all together. The post-processing should be able to deal with that, as the code to write in these numbers needs yet to be written anyways.

asmaatamna: It was
```
logger_bbob_write_data(logger->fdata_file, 
    1, initial_solution_fvalue, initial_solution_fvalue, 
    logger->optimal_fvalue, inner_problem->initial_solution,
    problem->number_of_variables);
```
before. And since the test in line ```432``` is
```
if (logger->number_of_evaluations == 1 && problem->number_of_constraints > 0)
```
I assumed there should be ```0``` there. I think there was an issue with the first entry being non-zero. Maybe @brockho remembers better.

nikohansen: I remember there was an issue that the evaluation count in the first output line should/must be 1. As the constraints evaluation count column is newly introduced, there should be no constraint there to begin with, unless we constrain ourselves.

nikohansen: See also https://github.com/numbbo/coco/issues/1518#issuecomment-327840025

PR #1542, opened 2017-09-08
added data format and logger name in info files

asmaatamna:


code-experiments/src/logger_bbob.c
```diff
@@ -421,23 +443,26 @@ static void logger_bbob_evaluate(coco_problem_t *problem, double *x, double *y)
        * does not correspond to evaluation counter 1.
        */
       logger_bbob_write_data(logger->fdata_file, 
-          1, initial_solution_fvalue, initial_solution_fvalue, 
-          logger->optimal_fvalue, inner_problem->initial_solution, problem->number_of_variables);
+          1, 0, initial_solution_fvalue, initial_solution_fvalue,
```
brockho: That certainly does not seem correct, see also the previous comments here: https://github.com/numbbo/coco/pull/1537#discussion_r137581976

The original idea of this part (within the if of line 432) was that in case the first f-evaluation is coming after the first g-evaluations, that we ensure an entry with 1 (f+g) evaluations by writing the feasible solution provided by the problem at the first g-evaluation (and hence f+g evaluation of 1 as well). Hence, we should write here at least
```
0, 1, initial_solution_fvalue, initial_solution_fvalue,
```
or remove this entire if clause if we can assume that the postprocessing will not choke on a missing g-evaluation entry of 1 (e.g. if we later on only want to display g-evaluations in the postprocessing).

code-experiments/src/logger_bbob.c
```diff
@@ -421,23 +443,26 @@ static void logger_bbob_evaluate(coco_problem_t *problem, double *x, double *y)
...
+          logger->optimal_fvalue, inner_problem->initial_solution,
+          problem->number_of_variables, logger->constrained_problem);
       coco_observer_targets_trigger(logger->targets, initial_solution_fvalue - logger->optimal_fvalue);
       logger_bbob_write_data(logger->tdata_file, 
-          1, initial_solution_fvalue, initial_solution_fvalue,
-          logger->optimal_fvalue, inner_problem->initial_solution, problem->number_of_variables);
+          1, 0, initial_solution_fvalue, initial_solution_fvalue,
```
brockho: [the same comment as above applies here]

code-experiments/src/logger_bbob.c
nikohansen: I am not sure I am in favor of writing lines with evaluation count zero, unless I understand why this would be helpful afterwards.

Issue #1496, opened 2017-08-25
warning in pprldmany with function=<closed file...


nikohansen:
```
cocopp/compall/pprldmany.py:823: UserWarning: The data for algorithm ./exdata/fmin_on_bbob_budget0002xD-001 and function <closed file 'ppdata/rando_fmin/pprldmany_03D_separ.tex', mode 'w' at 0x110bb9a50> are missing
  warnings.warn('The data for algorithm %s and function %s are missing' % (alg, f))
```
is a bug, because `f` is not the right variable to print here.

islandblue:
Earlier in the code, f is assigned what I believe to be a function number/index, but before this line, it is reassigned a file object. Sadly, I wasn't able to figure out what the correct expression should be, due to lack of knowledge on the data structures involved. I'll leave this to someone more familiar with the code in question.

nikohansen:
Using a different variable name for the file should probably do then.

brockho:
@islandblue remarks that the warning is issued much after the function `f` is defined (within an already finished loop) and we might also consider moving the warning.

brockho:
Seems to be closed with #1547 already.

PR #1547, opened 2017-09-11
Use different name for file object #1496

islandblue:
Changed `f` to `file_obj`, as discussed in #1496


nikohansen:
Any idea why we see ``Validation failed! File latex_commands_for_html.html should be regenerated!`` in the second and third test that don't run through?

brockho:
I thought I knew what was going on, but now not anymore :-(

Normally, the above message is shown when the `latex_commands_for_html.html` file got outdated. We then have to generate it again with the LaTeX to HTML converter via `python preparehtml.py`. However, when I do this now, there is no change in the file (besides a new date). The error must therefore come from something else. @dtusar: any idea?

dtusar:
I noticed that the generated `html` contains a string which does not exist in the code any more. Perhaps it does not use the latest version of `cocopp`. I have to check this on the slaves.

nikohansen:
Changes merged with ee01661b

PR #1549, opened 2017-09-12
Changed the way how to get the package path.

dtusar:


nikohansen:
@dtusar could you check why `ci/circleci: test_ubuntu_latest_python2` fails? It is not supposed to be broken (did work before) https://circleci.com/gh/numbbo/coco

Issue #169, opened 2015-11-23
observers/loggers should take recommendations into account


nikohansen:
By default, all observers/loggers should log the recommendation calls. They log the same information as for a function call for the recommended solution, without increasing the function evaluation counter.

ttusar:
How should the logger treat the non-existing y values of the recommended solutions?

nikohansen:
It says "they log the same information as for a function call". What is the question then? What are the possible alternatives to choose from?

nikohansen:
The proposition was that, by default, another line is added to the output which records the values for the recommended solution (same format as for a function call), while using the same number of evaluations (i.e. time stamp) as for the last recorded entry (i.e. evaluation or recommended solution).

ttusar:
Sorry for that question, everything is clear now.

We have just discussed recommendations with @brockho and came to the following idea that covers many possible uses for recommendations. For example, the recommendations could be used for recommending solutions for the noisy problems, or if the algorithm uses some kind of model of the function it is optimizing, but also for giving the user the possibility to log more than what the chosen logger does (the user might want to output the current population at each generation, for example, and right now it can't do any such thing).

So, the idea is to log recommendations into separate files (different from the ones used for the regular logging). In addition to the evaluation number, y and x values, we could also log an additional string through which the user can output some additional information for each evaluation (we would need to change the interface a bit). We wouldn't set a limit to the number of recommendations the user wishes to log (more than one recommendation can be done simultaneously) and we would leave it to the post processing to interpret them.

nikohansen:
I don't think we should (ab)use recommendations for anything else than what they are named for: to separate solution for which the solver wants to know a function value from solutions which the solver recommends as best current approximation to the optimum. Abusing this interface is on the long run quite likely to be more confusing than helpful and bug-prone. If the user wants to log something out of the ordinary which the logger can actually log, then the logger should "provide" a generic interface for doing so, e.g. a simple log-call.

ttusar:
Good point, it would probably be better to provide a separate interface...

ttusar:
Still, for the MO case, it makes sense to log the recommendations into a separate file (in the MO case you can decide to log the nondominated solutions only at the end or even to not log them at all).

ttusar:
[...]

PR #1558, opened 2017-09-14
Added non-anytime example experiment in Python.

ttusar:


code-experiments/build/python/example_experiment_non_anytime.py
```diff
@@ -0,0 +1,165 @@
...
+    # a list of increasing budgets (to be additionally multiplied by dimension)
+    # gradually increase `max_budget` to 10, 1000, ... or replace with a user-defined list
+    budget_multiplier_list = default_budget_list(max_budget=5)
+
+    ### prepare (see http://numbbo.github.io/coco-doc/C/ for suite and observer parameters)
+    suite = cocoex.Suite(suite_name, "", "")
+    observer = cocoex.Observer(suite_name, "result_folder: {} algorithm_name: {}".format(output_folder, algorithm_name))
+
+    ### go
+    for index, _ in enumerate(suite):
```
nikohansen: shouldn't this read rather 
```python
for index in range(len(suite)):
```

code-experiments/build/python/example_experiment_non_anytime.py
```diff
@@ -0,0 +1,165 @@
...
+        with ProblemNonAnytime(suite, observer, index) as problem:
+            problem.print_progress()
+            for budget in [problem.dimension * x for x in budget_multiplier_list]:  # iterate over budgets
+                problem.reset()
+                x0 = problem.initial_solution  # initial solution
+                # apply restarts while neither the problem is solved nor the budget is exhausted
+                while problem.evaluations < budget and not problem.final_target_hit:
+                    fmin(problem, x0, solver, budget)
```
nikohansen: AFAICS the budget passed to the solver needs to be recomputed for each restart. Passing `budget - problem.evaluations` to the solver seems to be the right choice.

code-experiments/build/python/example_experiment_non_anytime.py
```diff
@@ -0,0 +1,165 @@
...
+                    x0 = problem.random_solution  # a random initial solution for restarted algorithms
+            problem.free()
```
nikohansen: This line seems superfluous, because here problem.__exit__() is called anyways. That is kind-of the point of the `with` construct.

code-experiments/build/python/example_experiment_non_anytime.py
```diff
@@ -0,0 +1,165 @@
...
+import os
+import webbrowser
+
+# for solvers
+import scipy.optimize
+import cma
+
+
+def default_budget_list(max_budget=10, num=100):
+    """Produces a budget list with at most `num` different increasing budgets within [1, `max_budget`] that are equally
```
nikohansen: Python lines should generally not be longer than 80 characters see https://www.python.org/dev/peps/pep-0008/#maximum-line-length. Same for above docstring, 80-character lines are nicer to read anyways.

code-experiments/build/python/example_experiment_non_anytime.py
```diff
@@ -0,0 +1,165 @@
...
+    spaced in the logarithmic space.
+    """
+    from math import log10
```
nikohansen: I believe that local imports are frowned upon in Python. I would tend to say rather
```python
import numpy as np
```
on top and living with 
```python
np.unique(np.logspace(0, np.log10(max_budget), num=num).astype(int))
```
is better style in an offline script.

code-experiments/build/python/example_experiment_non_anytime.py
```diff
@@ -0,0 +1,165 @@
...
+                    fmin(problem, x0, solver, budget)
+                    x0 = problem.random_solution  # a random initial solution for restarted algorithms
+            problem.free()
+
+    ### post-process data
+    cocopp.main(observer.result_folder)  # re-run folders look like "...-001" etc
+    webbrowser.open("file://" + os.getcwd() + "/ppdata/index.html")
+
+
+class ProblemNonAnytime(object):
```
nikohansen: Given there is no way to prevent dependency on `cocoex` here, I don't see a reason why this class shouldn't become part of the `cocoex` module.

code-experiments/build/python/example_experiment_non_anytime.py
nikohansen: We should be aware of the point that there is no guaranty that the non-anytime algorithm _actually evaluates_ the solution it proposes as solution to the problem after the budget is exhausted. Therefore, arguably, we need to evaluated the return value of the solver too. Ideally we only do this if it hasn't done by the solver itself.

nikohansen: This comment also is a reminder of the (open) recommendation question #169. 

Here, IMHO we need to make sure that the return value of the solver is taken into account for performance evaluation.

First, simple fix: just evaluate the returned solution.

Second, I am contemplating over a `RecommendationHandler` class, which addresses #169 in some way.

ttusar: Good catch, thanks!

code-experiments/build/python/example_experiment_non_anytime.py
ttusar: I completely agree. We should also consider moving some of the classess/functions from `example_experiment.py` into `cocoex` too...

I'm leaving it in for the next commit. Will you move it to `cocoex` or should I?

code-experiments/build/python/example_experiment_non_anytime.py
ttusar: True! I was mislead by PyCharm... (https://stackoverflow.com/questions/26808681/why-does-pycharm-use-120-character-lines-even-though-pep8-specifies-79)

The PEP 8 guide is actually stricter: *Limit all lines to a maximum of **79 characters**. 
For flowing long blocks of text with fewer structural restrictions (docstrings or comments), the line length should be limited to **72 characters**.*

I'm now using a limit of 79 characters for everything, but the code is less readable now (in my opinion, of course). I can shorten the docstrings/comments to 72 characters, if you think it's necessary.

BTW, the guide says also: *For code maintained exclusively or primarily by a team that can reach agreement on this issue, it is okay to increase the nominal line length from 80 to **100 characters** (effectively increasing the maximum length to 99 characters), provided that comments and docstrings are still wrapped at 72 characters.* Just a hint ;-)

code-experiments/build/python/example_experiment_non_anytime.py
ttusar: Should we equate non-anytime algorithms with surrogate-assisted algorithms (I guess this is what is the underlying motivation here)? The same care should be taken in the anytime scenario as well, shouldn't it?

ttusar:
How does this work now that the improved code is ready? Should I close this pull request and make a new one or is it possible to update this one in some way?

code-experiments/build/python/example_experiment_non_anytime.py
```diff
@@ -0,0 +1,193 @@
...
+
+    # a list of increasing budgets (to be additionally multiplied by dimension)
+    # gradually increase `max_budget` to 10, 100, ...
+    # or replace with a user-defined list
+    budget_multiplier_list = default_budget_list(max_budget=5)
+
+    ### prepare (see http://numbbo.github.io/coco-doc/C/ for parameters)
+    suite = cocoex.Suite(suite_name, "", "")
+    observer = cocoex.Observer(suite_name,
+                               "result_folder: {} algorithm_name: {}".format(
```
nikohansen: For the record, this syntax is not compatible with Python 2.6, where you would need to specify the argument ID with the `{}` brackets, like ``"result_folder: {0} algorithm_name: {1}".format(``.

ttusar:
I will merge this now. Additional changes (if needed) can always be made later.

code-experiments/build/python/example_experiment_non_anytime.py
nikohansen: I don't equate non-anytime with surrogate-assisted. Yes, recommendations should be available always. They aren't yet. I see this as a minor issue in the noiseless anytime scenario.

ttusar: Why is this a big issue here?

code-experiments/build/python/example_experiment_non_anytime.py
nikohansen: The `example_experiment.py` has been cleared from the helper code. Currently I am not sure where the `ProblemNonAnytime` class should go. It seems we should have another file under the python folder.

code-experiments/build/python/example_experiment_non_anytime.py
nikohansen: Because the algorithm is non-anytime, I don't expect it (somehow by definition) to produce good solutions before it returns one.

ttusar: The latest commit takes care of this. I closed all non-anytime-pertaining issues, please reopen if needed.

code-experiments/build/python/example_experiment_non_anytime.py
ttusar: Oh, I thought it will go into the `cocoex` module... Is there any reason why it shouldn't be put there?

PR #1628, opened 2017-10-17
Biobj expose reference

ttusar:
Exposed the largest f-values of interest in `coco.h` and created interfaces in Python and Java

### DIFF
--- a/code-experiments/build/python/cython/interface.pyx
+++ b/code-experiments/build/python/cython/interface.pyx
@@ -679,11 +679,7 @@ cdef class Problem:
                 self._lower_bounds[i] = coco_problem_get_smallest_values_of_interest(self.problem)[i]
             if coco_problem_get_largest_values_of_interest(self.problem) is not NULL:
                 self._upper_bounds[i] = coco_problem_get_largest_values_of_interest(self.problem)[i]
-        ## FIXME: If the code above is inefficient, so is this one:
-        self._largest_fvalues_of_interest = np.inf * np.ones(self._number_of_objectives)
-        if self._number_of_objectives > 1 and coco_problem_get_largest_fvalues_of_interest(self.problem) is not NULL:
-            for i in range(self._number_of_objectives):
-                self._largest_fvalues_of_interest[i] = coco_problem_get_largest_fvalues_of_interest(self.problem)[i]
+        self._largest_fvalues_of_interest = None
         self.initialized = True
         return self
     def constraint(self, x):
@@ -880,6 +876,11 @@ cdef class Problem:
     def largest_fvalues_of_interest(self):
         "largest f-values of interest (defined only for multi-objective problems)"
         assert(self.problem)
+        if self._number_of_objectives > 1 and coco_problem_get_largest_fvalues_of_interest(self.problem) is not NULL:
+            # Not a particulary nice way of doing this, mostly due to cython
+            self._largest_fvalues_of_interest = np.inf * np.ones(self._number_of_objectives)
+            for i in range(self._number_of_objectives):
+                self._largest_fvalues_of_interest[i] = coco_problem_get_largest_fvalues_of_interest(self.problem)[i]
         return self._largest_fvalues_of_interest
 
     def free(self, force=False):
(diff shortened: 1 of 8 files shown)


ttusar:
Ah, I inadvertently merged this (it was supposed to wait here for a review)... Should I revert?

code-experiments/build/python/cython/interface.pyx
```diff
@@ -869,6 +872,16 @@ cdef class Problem:
     def best_observed_fvalue1(self):
         assert(self.problem)
         return coco_problem_get_best_observed_fvalue1(self.problem)
+    @property
+    def largest_fvalues_of_interest(self):
+        "largest f-values of interest (defined only for multi-objective problems)"
+        assert(self.problem)
+        if self._number_of_objectives > 1 and coco_problem_get_largest_fvalues_of_interest(self.problem) is not NULL:
+            # Not a particulary nice way of doing this, mostly due to cython
+            self._largest_fvalues_of_interest = np.inf * np.ones(self._number_of_objectives)
```
brockho: Not clear why we need the `np.inf` here...

code-experiments/src/coco_problem.c
```diff
@@ -523,6 +523,15 @@ const double *coco_problem_get_largest_values_of_interest(const coco_problem_t *
   return problem->largest_values_of_interest;
 }
 
+const double *coco_problem_get_largest_fvalues_of_interest(const coco_problem_t *problem) {
```
brockho: Shouldn't the returned nadir point be a vector of length #objectives? Why do we return a double only?

code-experiments/build/python/cython/interface.pyx
ttusar: We don't really, a `np.zeros` would probably suffice...

code-experiments/src/coco_problem.c
ttusar: We return `const double *`

brockho: Ah, sorry. Didn't see it. You are right.

code-experiments/build/python/cython/interface.pyx
nikohansen: Lines 880-883 could (should?) be replace by
```python
# calling coco_... several times should not be necessary, but it's not clear how to avoid it
self._largest_fvalues_of_interest = np.asarray(
        [coco_problem_get_largest_fvalues_of_interest(self.problem)[i]
         for i in range(self._number_of_objectives)])
```
This also eliminates the misleading initialization.

code-experiments/build/python/cython/interface.pyx
```diff
@@ -636,6 +637,7 @@ cdef class Problem:
     # cdef public np.ndarray upper_bounds
     cdef public np.ndarray _lower_bounds
     cdef public np.ndarray _upper_bounds
+    cdef public np.ndarray _largest_fvalues_of_interest
```
nikohansen: I know it's just copy-paste, but the public qualifier doesn't look "right" for an attribute whose name starts with an underscore. Maybe it is historical left-over for the `_boundary...` attributes.

ttusar: It is, sorry.

code-experiments/build/python/cython/interface.pyx
ttusar: Yes, this should do it...

code-experiments/build/python/cython/interface.pyx
ttusar: Seems to work if I remove the `public` keyword. Should I do this for `_lower_bounds` and `_upper_bounds` too?

ttusar: Some more information regarding public attributes (from http://cython.readthedocs.io/en/latest/src/reference/extension_types.html):

> To make attributes accessible to Python, they must be declared `public` or `readonly`

Since we don't need to access any of these three attributes starting by an underscore from Python, it should be okay to remove the `public` keyword for all three of them.

Issue #1623, opened 2017-10-17
do.py verbose


nikohansen:
```python 
./do.py verbose build-c
```
should be the default and quiet should be an option.

PR #1634, opened 2017-10-20
Make verbose default for c builds, add quiet option (#1623)

islandblue:
Plus some make-up on do.py. Should still be easy enough to distinguish which changes are related to the issue.

### DIFF
--- a/do.py
+++ b/do.py
@@ -45,6 +45,8 @@ MATLAB_FILES = ['cocoCall.m', 'cocoEvaluateFunction.m', 'cocoObserver.m',
                 'cocoSuiteGetNextProblem.m', 'cocoSuiteGetProblem.m']
 
 _verbosity = False
+# Do not suppress build messages unless specifically requested
+_build_verbosity = True
 
 ################################################################################
 ## C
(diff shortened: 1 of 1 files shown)


do.py
```diff
@@ -274,8 +295,10 @@ def install_postprocessing(package_install_option = []):
     expand_file(join('code-postprocessing', 'setup.py.in'),
                 join('code-postprocessing', 'setup.py'),
                 {'COCO_VERSION': git_version(pep440=True)})
-    # copy_tree('code-postprocessing/latex-templates', 'code-postprocessing/cocopp/latex-templates')
-    python('code-postprocessing', ['setup.py', 'install'] + package_install_option, verbose=_verbosity,
+    # copy_tree('code-postprocessing/latex-templates',
+    #           'code-postprocessing/cocopp/latex-templates')
```
nikohansen: One of the cases where obeying PEP 8 makes the code less readable, I would say.

do.py
```diff
@@ -831,6 +855,11 @@ def verbose(args):
     main(args)
     _verbosity = False
 
+def quiet(args):
```
nikohansen: As a general rule, any function should have a docstring. Here, as single line would suffice, but it should be there.

nikohansen:
I don't understand why the appveyor test is labeled as failed? Otherwise the pull can be merged.

islandblue:
Thank you @nikohansen, I added the docstrings and tried to exercise common sense to choose which lines to unwrap and which to keep.

do.py
```diff
@@ -175,14 +170,10 @@ def build_c_integration_tests():
...
                 {'COCO_VERSION': git_version(pep440=True)})
-    copy_file('code-experiments/src/bbob2009_testcases.txt',
-              'code-experiments/test/integration-test/bbob2009_testcases.txt')
-    copy_file('code-experiments/src/bbob2009_testcases2.txt',
-              'code-experiments/test/integration-test/bbob2009_testcases2.txt')
-    make("code-experiments/test/integration-test", "clean",
-         verbose=_build_verbosity)
-    make("code-experiments/test/integration-test", "all",
-         verbose=_build_verbosity)
+    copy_file('code-experiments/src/bbob2009_testcases.txt', 'code-experiments/test/integration-test/bbob2009_testcases.txt')
```
nikohansen: sorry, my comment was regarding PEP 8 and _incommented_ code, otherwise line breaks are OK

Issue #1638, opened 2017-10-21
the master branch has become unusable with the latest version of matplotlib


nikohansen:
see #1636 and https://github.com/numbbo/coco/issues/1481#issuecomment-336171594

nikohansen:
As a fix we may cherrypick the change of commit 7692121decc55f49afa1a1e95d10d9237d901e49 into the master branch.

nikohansen:
I did the cherrypick, but due to our testing logic it seems difficult to push changes to the master branch.

brockho:
It seems that our previous proposal of having a release branch wasn't the worst of all ideas. The master branch should then be protected in a way that tests on this release branch need to pass before the merge and not anymore from development. This will allow the two use cases of quick fixes in the master and larger releases from development. Or am I wrong?

nikohansen:
> It seems that our previous proposal of having a release branch wasn't the worst of all ideas. 

Exactly what I was thinking.

>  This will allow the two use cases of quick fixes in the master and larger releases from development.

I don't think that is yet the ideal solution. I don't want to hurry a release, just because it is "impossible" to push a small fix to the master.

Protecting the master by a test on one or several of the test branches should do the trick either AFAICS.

brockho:
We implemented now the last suggestion of @nikohansen above (protecting the master via `devel-test1` and `test-nightly`. The issue can thus be closed.

PR #1639, opened 2017-10-21
trying to fix incompatibility with matplotlib version 2.0.2 -> 2.1

nikohansen:
thereby solving issue #1638

### DIFF
--- a/code-postprocessing/cocopp/__main__.py
+++ b/code-postprocessing/cocopp/__main__.py
@@ -271,7 +271,7 @@ def main(args):
         
         # testing data from bbob-noisy suite:
         t0 = time.time()
-        result = os.system(python + command +
+        result = os.system(python + command + ' --omit-single ' +
                            join_path(data_path, 'MCS_huyer_noisy.tgz') +
                            join_path(data_path, 'BFGS_ros_noisy.tgz'))
         print('**  subtest 9 finished in ', time.time() - t0, ' seconds')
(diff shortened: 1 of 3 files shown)


nikohansen:
Unfortunately, I don't know how to run a pull-request on the Jenkins nightly build branch.

dtusar:
We currently don't have the nighty build for the master branch but I can prepare one.

nikohansen:
@dtusar, do you understand the problem? To my understanding, the tests must pass _before_ a branch can be pushed to master. I don't exactly see how a nightly build on the master branch would help to solve this problem nor why we would need one.

brockho:
Would it be enough to equip one of the `devel-test` branches with the functionality to call the nightly build by hand in Jenkins? If the answer is yes, I can set this up rather quickly I would guess.

dtusar:
The other option is to have a `test-nightly` branch and every time you make a push to this branch the nightly tests automatically start.

nikohansen:
Isn't the problem that the nightly build, when triggered, pulls from the development branch? If I could push the pull request branch to the nightly build (which is the test that the master branch requires), the problem would be solved.

Setting up a `test-nightly` should work if also the requirements of the master branch are changed, which seems a possible solution.

dtusar:
I created the `test-nightly` branch (from master) and added two new jobs (`Coco-test-nightly` and `Coco-test-nightly-win`) to Jenkins that start when you push something to this branch. I hope this helps.

nikohansen:
FTR, the current master branch fails when tested on the devel-test2 branch.

brockho:
Seems as if the latest renaming of data files causes the problem when running the tests on the master branch: the former `bbob-biobj/2016/RS-4.tgz` is now called `bbob-biobj/2016/RANDOMSEARCH-4_Auger_bbob-biobj.tgz`.

PR #1648, opened 2017-10-24
Use latest version of matplotlib

islandblue:


islandblue:
Failing appveyor tests to be discussed and handled in #1651

PR #1673, opened 2017-11-26
Check length before indexing

islandblue:


islandblue:
`verify-postprocessing` fails on Jenkins Windows machine.

> Original line:         (f+g)-evaluations in any trial. Notched boxes: interquartile range with median of simulated runs; 
> 
> Generated line:         f-evaluations in any trial. Notched boxes: interquartile range with median of simulated runs; 

Seems to me like the fix we discussed in today's meeting, i.e. the correction of caption text to match the plot, also requires the expected output of the tests to be updated accordingly. Barring additional failures, I intend to merge in spite of this failure, on the grounds that it is clearly unrelated to the changes in this pull request.

Issue #1689, opened 2017-12-22
significant digits in html tables


nikohansen:
The number of evaluations in the second to last column in the `pptable.html` files is written with two digits of precision even when the `%d` format is used. 
<img width="382" alt="screen shot 2017-12-22 at 22 10 24" src="https://user-images.githubusercontent.com/7316439/34312515-0e103490-e765-11e7-9509-892439baf3fd.png">
This should be considered as a bug. I suspect that `writeFEvalsMaxPrec` is used where `writeFEvals2` would be the better choice.

nikohansen:
Addressed with #1690

PR #1690, opened 2017-12-22
small format adjustment addressing #1689

nikohansen:
Now the maximal width is defined as 3 chars, which cancels the precision prescription. The latter lead to irritating rounding results as shown in #1689

### DIFF
--- a/code-postprocessing/cocopp/pptable.py
+++ b/code-postprocessing/cocopp/pptable.py
@@ -293,8 +293,8 @@ def main(dsList, dims_of_interest, outputdir, latex_commands_file):
                     tableentryHtml = writeFEvalsMaxPrec(tmp, 2)
 
                     if np.isinf(tmp) and i == len(data)-1:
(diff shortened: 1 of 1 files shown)


kostasvar:
Some results with the _old_ code : 


screen shot 2017-12-23 at 18 58 39


and with the new : 


screen shot 2017-12-23 at 18 59 34


It seems that the accuracy issue is fine now. The dispersion measure in braces seems to change significantly though (sometimes reduced), and I would check if this is normal.

kostasvar:
There are still rounded results when running the post-processing in some cases. I get the following when running the post-processing for the results of two algorithms (large-scale, 0100xD):
screen shot 2017-12-27 at 18 21 37

nikohansen:
If you run it again, do you get the same numbers again?

kostasvar:
No, they are different every time!

nikohansen:
OK, I did the changes, no need to make a new pull request then.

PR #1694, opened 2017-12-27
Fixed coco_create_directory and coco_create_unique_directory

ttusar:


code-experiments/src/coco_utilities.c
```diff
@@ -156,57 +157,72 @@ static int coco_file_exists(const char *path) {
...
 
-  tmp = coco_strdup(path);
-  /* Remove possible trailing slash */
+  /* Remove possible leading and trailing (back)slash */
   if (tmp[len - 1] == path_sep)
     tmp[len - 1] = 0;
-  for (p = tmp + 1; *p; p++) {
+  if (tmp[0] == path_sep)
+    tmp++;
+
```
brockho: Am I correct that this will not allow to use absolute paths under *nix systems? What if I want to write to `/home/dimo/`? In which use case is it important to remove the leading slash?

ttusar: Unless I'm mistaken, this was always meant to work only for relative paths. It's true, though, that this should be explicitly written, and it's not now.

brockho:
The Jenkins issue might be again related to a github API change that does not fit with our Jenkins version (2.92). I started to install version 2.98 on the qualification server to test it before to migrate our production server.

ttusar:
Summarizing the latest tests for this pull request:
- AppVeyor test on Python 2.7 failed because "Build execution time has reached the maximum allowed time for your plan (60 minutes)."
- The Jenkins leak-check test on Mac failed, but it's unclear why (the console output after my changes (https://ci.inria.fr/numbbo/job/Coco-test2-mac/dotype=leak-check,label=numbbo-osx10p12/299/console) is the same as the console output on devel-test1 after some changes by @nikohansen  (https://ci.inria.fr/numbbo/job/Coco-test1-mac/dotype=leak-check,label=numbbo-osx10p12/507/console)
- All other tests passed, however this is not visible from the summary here at Github (all Jenkins tests are shown as pending)

Issue #1698, opened 2018-01-11
run-python fails silently


ttusar:
Calling `python do.py run-python` fails without raising an exception, so it goes unnoticed. 

The problem is that the `example_experiment.py` script requires an additional parameter - the name of the suite, which is missing in `do.py`. The fix for this particular problem is easy - line 346 of `do.py` needs to be changed from
`               ['example_experiment.py'])`
to
`               ['example_experiment.py', 'bbob'])`

However, the more general problem of silent fails also needs to be addressed at some point (not sure whether we already have any open issues on this matter).

BTW, I have tried this on the `development` branch.

ShubhGupta2125:
I am a beginner and would like to work on this issue. Can i make the change that you have mentioned above?

nikohansen:
You can make the change after you have forked the code (either online or offline). Use the development branch for that. If you try to change the code online, I think at some point the fork will be done automatically, because you are not allowed to change the original code. After that, test the changes via `do.py`. Then you can commit a pull request with the new code to the development branch.

ShubhGupta2125:
@nikohansen  please review it.

dtusar:
@brockho Can this issue be closed?

brockho:
Given what @ttusar wrote above:

"However, the more general problem of silent fails also needs to be addressed at some point (not sure whether we already have any open issues on this matter)."

I guess, the issue is still open.

dtusar:
But is this still something that needs to be done for the next release milestone?

brockho:
No. I removed the milestone from this issue.

olafmersmann:
Closed in anticipation of the 2.6.3 release that replaces `do.py`.

PR #1701, opened 2018-01-18
#1698 

ShubhGupta2125:
#1698 added the name of suite parameter in run_python in do.py


nikohansen:
Did you see the conflict in ``cocopp/__main__.py``?

nikohansen:
The changes of the pull request are probably based on the original master branch instead of the development branch, hence we see the conflict.

PR #1705, opened 2018-01-30
Pullrequest 1694

brockho:
This is the same than pull request #1694 which was coming from the devel-test2 branch and was stored temporarily in the pullrequest-1694 branch to free the test branch.

If not mistaken, some of the tests have not gone through on this code yet.


nikohansen:
I restarted two of the tests, let's see whether they run through this time.

Issue #1670, opened 2017-11-23
Fixed-target ECDF plots broken for bbob-biobj


brockho:
When postprocessing a recently produced data set (NSGA-II from [platypus](https://github.com/Project-Platypus/Platypus) with a budget of only 10n function evaluations, see [exdata-platypus-NSGA-II-m100p100-budget10D.zip](https://github.com/numbbo/coco/files/1500255/exdata-platypus-NSGA-II-m100p100-budget10D.zip)), and comparing it with the NSGA-II from the data-archive, I got very strange fixed target ECDFs such as 

![pprldistr_05d_all](https://user-images.githubusercontent.com/13706996/33189707-b9675cb4-d0a5-11e7-91e8-f92e66c67945.png)

[aggregation over all functions in 5-D]

Strange:
* the graph for one algorithm (according to the caption, the platypus one) going up that directly to 1.0 right at 20n function evaluations
* the markers for the other algorithm

nikohansen:
The markers seems to suffer from some sort of index shift by one: e.g., the first marker is placed on its own y-value and on the y-value of the second marker...

brockho:
When we looked again at the plot (and only the plot), we realized that for the platypus algorithm no target was actually hit: the line on the left with the circled markers is very likely the max-budget line and not a data line.

Maybe, the wrong markers come from the fact that one data for one algorithm is empty here.

islandblue:
I failed to reproduce this plot. I get the following:

![pprldistr_05d_noiselessall](https://user-images.githubusercontent.com/27438398/35047255-a2f0134e-fb99-11e7-8767-08f6cf9c7179.png)

I got this by running `python -m cocopp exdata NSGA-II` where exdata is the contents of the archive @brockho linked and `NSGA-II` resolves to `NSGA-II-MATLAB_Auger_bbob-biobj`. cocopp is the latest commit to the development branch ([`d970bffe`](https://github.com/numbbo/coco/tree/d970bffe06a953c585f9424a7a6121cbc1520467)) and I used Python 2.7.

@brockho, if you used a different method, I'd like to retrace your steps.

islandblue:
It occurred to me to try `python -m cocopp NSGA-II exdata` just in case order matters. Same result though.

islandblue:
Post-processed the same data 20 times, in case the bug is RNG-related. Still can't get a buggy graph.

brockho:
[...]

PR #1710, opened 2018-02-02
Fix for issue #1670

islandblue:
Erroneous markers in fixed-target ECDF

### DIFF
--- a/code-postprocessing/cocopp/ppfig.py
+++ b/code-postprocessing/cocopp/ppfig.py
@@ -545,25 +545,29 @@ def plotUnifLogXMarkers(x, y, nbperdecade, logscale=False, **kwargs):
 
     This function only works with monotonous graph.
     """
-    res = plt.plot(x, y, **kwargs)  # shouldn't this be done in the calling code?
+    line_args = kwargs.copy()
+    line_args['marker'] = ''
+    line_args['label'] = ''
+    res = plt.plot(x, y, **line_args)  # shouldn't this be done in the calling code?
 
     if 'marker' in kwargs and len(x) > 0:
         # x2, y2 = downsample(x, y)
         x2, y2 = marker_positions(x, y, nbperdecade, 19, plt.axis(),
                                   np.log10 if logscale else None)
-        res2 = plt.plot(x2, y2)
-        for i in res2:
-            i.update_from(res[0])  # copy all attributes of res
-        plt.setp(res2, linestyle='', label='')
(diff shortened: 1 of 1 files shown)


islandblue:
The commit tested green everywhere.

code-postprocessing/cocopp/ppfig.py
```diff
@@ -545,25 +545,29 @@ def plotUnifLogXMarkers(x, y, nbperdecade, logscale=False, **kwargs):
...
                                   np.log10 if logscale else None)
-        res2 = plt.plot(x2, y2)
-        for i in res2:
-            i.update_from(res[0])  # copy all attributes of res
-        plt.setp(res2, linestyle='', label='')
+        marker_args = kwargs.copy()
+        marker_args['drawstyle'] = 'default'
+        marker_args['linestyle'] = ''
+        marker_args['label'] = ''
+        res2 = plt.plot(x2, y2, **marker_args)
```
dtusar: I noticed that in the new code the `drawstyle` is set to 'default' (for `res2`) but in the old code it wasn't. Similarly below `linestyle` and `marker` are set to '' (for `res3`). Was this intended?

islandblue: Indeed, all intended.

The first plot is to draw the line, the second is to place the markers and the third is to add the label. In `kwargs`, `linestyle`, `marker` and `label` are provided to style each of these plot elements, respectively. For the first plot, `linestyle` matters but `marker` and `label` must be unset. For the second plot, `marker` matters but `linestyle` and `label` must be unset. For the third plot, `label` matters and since it's a plot of `NaN` vs `NaN`, `linestyle` or `marker` are supposed to be ineffectual. However, since making a plot invisible by passing `NaN` for coordinates is relying on exceptional handling of this special value internal to `matplotlib` and the same effect could be accomplished by unsetting `linestyle` and `marker` in a more self-explanatory fashion, I thought I'd add the additional code in. After all, passing empty lists as coordinates also worked until recently. Better to not have to keep in mind or rely on `matplotlib`'s handling of `NaN` values.

Setting `drawstyle` to `default` for the marker plot _is_ the actual fix. When set to `steps`, `matplotlib` places extra points so that the plot looks like a staircase. That may work with the line but surely not with the way post-processing uses markers.

dtusar: Thanks for the clarification. Then it's fine for me, it can be merged.

Issue #1695, opened 2017-12-28
Best_parameter wrong for Katsuura function (f23)


brockho:
Following, #1672, the pull request #1679 and the discussions around these issues, we have to still correct the best_parameter for the Katsuura function (`f_23` in `bbob`). 

The problem stems from the fact that when updating the best_parameter in transform_vars_affine.c, we assume for the moment that the transformation's matrix is orthogonal which is not the case for the Katsuura function (where the matrix $M$ equals $Q\Lambda R$ with $Q$ and $R$ being any (pseudo-)random $n\times n$ matrix and $\Lambda$ a diagonal matrix).

This issue is critical because functions in the extended biobjective suite depend on it and reference archives and hypervolume values are, thus, wrongly computed at the moment.

Note that the test for equality of the `best_parameter` between the old python code of COCO and the current one is still missing, see also the discussion in #1679.

brockho:
The concrete issue with `f_23` is solved now, but at the moment, we have a problem with `f_16`, see also https://github.com/numbbo/coco/pull/1679.

brockho:
We agreed today that this issue can be closed as soon as pull request #1716 is accepted.

Issue #1666, opened 2017-11-16
Constrained Different Powers simpler than expected


brockho:
Looking at results from the preliminary `bbob-constrained` suite, we realized today with @asmaatamna and @nikohansen that the Different Powers problems are probably too simple (at least simpler than in the unconstrained case). The reason is that the constrained optimum lies typically in a too far distance to the unconstrained optimum such that the conditioning, an algorithm has to cope with, is rather low.

To improve on this, we should use a smaller shift/translation in the construction to choose the new optimum closer to the original one (rather in the order of `1e-5` than a fully random shift).

asmaatamna:
@nikohansen 's suggestion: a possible way to decide what value to choose: run CMA-ES on the function and see what the condition number is for a given distance to the optimum.

asmaatamna:
For the moment, the constrained optimum is generated such that its distance from the unconstrained optimum is ```1e-5``` (code in ```devel-asma```). The link below provides a ```jupyter notebook``` that runs cma on the bbob (unconstrained) different powers (for a given dimension/instance).

[https://drive.google.com/open?id=1rxSr3XLJDWHoby0XdHiJ7g-F3RG8PVQn](url)

brockho:
Do I interpret it correctly from the plots, that the condition number around the constrained optimum will be about `1e6` if we choose a distance of around `1e-5`?

asmaatamna:
@brockho yes.

nikohansen:
Where can I see the distance to the optimum in the plots?

asmaatamna:
@nikohansen sorry I didn't understand your question correctly. You can see the the difference between ```f``` and ```fopt``` on the usual ```cma``` ```f``` graphs (the algorithm is run on the "normalized" different powers). I didn't plot the distance to the optimum...

asmaatamna:
Upon looking at some graphs in ```5D```, it looks like a shift of ```1e-2``` is already sufficient (we may want to test other dimensions/instances however).

brockho:
Unfortunately, I cannot remember what was our reasoning on Jan 23 that 1e-2 as distance is already sufficient. Did we check this in other dimensions than 5? When I run the Jupyter notebook above in 40-D, I get the following:
![40d](https://user-images.githubusercontent.com/13706996/37203051-b6fd1f42-238c-11e8-9348-f004b1124333.png)

Where exactly do I see the 1e-2 now? And, again, why did we change this from 1e-5?

[...]

Issue #1637, opened 2017-10-21
optimal f-values in the constrained suite


nikohansen:
The function `bbob-constrained_f07_i03_d03`, for example, has an optimal f-value larger than `1e7`.  This seems to be numerically incompatible with our standard choice of target precisions going down to `1e-8`. That is, the disagreement between the mathematical model and the numerical model becomes relevant, which is a rather undesirable situation.

The preferable solution would be to change the function definition. Offsetting the function value is probably the simplest solution, but it has to be carefully reviewed for loss of precision. Scaling down the function value is another alternative.

nikohansen:
The distribution of `fopt`-values on `bbob-constrained`:
<img width="323" alt="screen shot 2017-10-22 at 14 40 03" src="https://user-images.githubusercontent.com/7316439/31863056-709fa55c-b748-11e7-97fa-fd0e3d9fc3e6.png">
Due to the construction, the minimal `fopt` values are close to -1000.

Scaling down the functions such that `fopt` values before adding the offset are not much larger than 1000 seems advisable.

nikohansen:
We may apply the following transformation:
```
f_opt_raw = f(x_opt)
f(x) -> f(x) / f_opt_raw^0.65 - f_opt_raw^0.35 +  f_bbob2009_compute_fopt(...)
```
with `f_bbob2009_compute_fopt(...)` in [-1000,1000].

asmaatamna:
After implementing the transformation above (changes in ```devel-asma```), the ECDFs of optimal function values are shown in figure below:
![ecdf_fopt_bbob-constrained](https://user-images.githubusercontent.com/14993138/34569834-416adc6e-f16a-11e7-82d2-910016d9f3ef.png)

The ```iPython notebook``` for plotting the ECDFs is available here: [https://drive.google.com/file/d/1xnJzkBAMamD0M-vAa7mxwUY2D1IR50MK/view?usp=sharing](url)

PR #1716, opened 2018-02-21
Devel asma

asmaatamna:


asmaatamna:
Issues addressed in this pull request are: #1666 , #1637 , #1695. I also recreated the data files so that the test-suites test runs again (with the data I recreated last time, the test was failing under Windows because of what seemed to be a numerical problem for a large vector X).

code-experiments/src/transform_obj_shift.c
```diff
@@ -48,21 +48,21 @@ static void transform_obj_shift_evaluate_function(coco_problem_t *problem, const
...
-  
+
   data = (transform_obj_shift_data_t *) coco_problem_transformed_get_data(problem);
   coco_evaluate_constraint(coco_problem_transformed_get_inner_problem(problem), x, y);
-  
+
   for (i = 0; i < problem->number_of_constraints; i++)
     y[i] += data->offset;
-}
+}*/
```
brockho: We should probably remove this part entirely instead of commenting it out to make the code cleaner overall and easier to read.

code-experiments/test/regression-test/create/create_suite_data.py
```diff
@@ -17,7 +17,7 @@
 import numpy as np
 import cocoex
 suite_names = cocoex.known_suite_names
-# suite_names = 'bbob-largescale'
+# suite_names = 'bbob-constrained'
```
brockho: This line can also be removed.

code-experiments/test/regression-test/test_suites.py
```diff
@@ -67,7 +67,7 @@ def regression_test_a_suite(suite_name, filename):
             try:
                 assert is_equal(f.constraint(x), xfc_dict[key][1])
             except AssertionError:
-                print(f.name, "index,x =", key, "stored x,f,con =",
+                print(f.name, "index,x =", key, "stored f,con =",
```
brockho: What is the reason for the disappearing `x` here? The `xfc_dict` was not changed in this pull request, or was it?

asmaatamna: The ```x``` was removed because ```xfc_dict[key]``` contains two values only: ```f``` and ```con```. Therefore, the printed string ```"stored x,f,con ="``` wasn't reflecting the actual printed values (```f```, ```con```). The ```xfc_dict``` was not changed, the ```x``` was a typo.

code-experiments/src/transform_obj_scale.c
```diff
@@ -0,0 +1,90 @@
...
+  size_t i;
+
+  if (coco_vector_contains_nan(x, coco_problem_get_dimension(problem))) {
+    coco_vector_set_to_nan(y, coco_problem_get_number_of_objectives(problem));
+    return;
+  }
+
+  data = (transform_obj_scale_data_t *) coco_problem_transformed_get_data(problem);
+  coco_evaluate_function(coco_problem_transformed_get_inner_problem(problem), x, y);
+
```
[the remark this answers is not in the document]
asmaatamna: Good point. If we decide to use ```problem->evaluate_function``` instead, then this should be done everywhere (```coco_evaluate_function``` is used in other transformations).

code-experiments/test/regression-test/test_suites.py
brockho: If this is the case, we should rename the `xfc_dict` to `fc_dict`, shouldn't we?

asmaatamna: The dictionary is indexed by keys of the form ```(index, x)```, maybe that's why there's an ```x``` in the name. Whatever we decide, the renaming is easy to do.

code-experiments/src/transform_obj_scale.c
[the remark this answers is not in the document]
nikohansen: If the evaluation number is increased and not reset, it is a bug and needs to be solved everywhere.

code-experiments/src/transform_obj_scale.c
```diff
@@ -0,0 +1,90 @@
...
+ * @brief Evaluates the transformed function.
+ */
+static void transform_obj_scale_evaluate_function(coco_problem_t *problem, const double *x, double *y) {
+  transform_obj_scale_data_t *data;
+  double *cons_values;
+  int is_feasible;
+  size_t i;
+
+  if (coco_vector_contains_nan(x, coco_problem_get_dimension(problem))) {
+    coco_vector_set_to_nan(y, coco_problem_get_number_of_objectives(problem));
```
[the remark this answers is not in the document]
nikohansen: It should be consistent with how it is done everywhere else at similar places.

PR #1721, opened 2018-02-28
Requesting a pull to numbbo:master from numbbo:development

nikohansen:


nikohansen:
It seems that pushing the branch to `devel-test1` wasn't the best of all ideas, because now `Coco-test1-mac/dotype=leak-check,label=numbbo-osx10p12` has failed and will prevent the merge of this commit into `master`.

Issue #1706, opened 2018-01-30
Triggers in the constrained casse


brockho:
After a discussion with @dtusar and @nikohansen last week, we came up with the following (had been already documented 1:1 in our private gitlab repository):

* we considered to use non-domination of $`(f, \sum g_i^+)`$ as trigger. This would suggest to use the number of exhausted trigger values as vertical steps in the ECDF. This is apparently pretty much the same as hypervolume when the objectives are taken in log scale. 
* it seems _considerably_ simpler to use $`f + \sum g_i^+`$ as trigger value. When we take the usual trigger values linear on the log scale, this seems to be very similar to the above approach (because $`\sum \log g = \log \prod g`$??). Remark that in both cases it is most effective to reduce the larger/largest value in the tuple or sum. That is, to move toward all values being equal, even without taking squares under the sum.

brockho:
Note that this issue will likely involve both the C logger/observer and the python postprocessing part of COCO. @asmaatamna agreed to look at the C logger/observer part.

brockho:
Current suggestion:
* first two columns as in current constraint format
* third column: current (f + sum of g_i)
* forth column: current f

We need to check, however, that this is backwards compatible, in particular wrt. the noisy case.

nikohansen:
FTR, only the third column would change in the above proposal.

brockho:
Correction: third column should always be the best (f + sum of g_i) [in order to be correct when triggering on other things that the third column, for example in the tdat files]

nikohansen:
@asmaatamna, I wonder whether the assignment of this issue to @islandblue was by mistake?

asmaatamna:
@nikohansen no, @islandblue volunteered to do the postprocessing part related to this issue (I'm guessing there would be some captions to adapt).

islandblue:
...as agreed on last week's meeting.

asmaatamna:
Just to be sure:
1. We only keep track of infeasible solutions that satisfy ```f - f(bestparameter) > 0```.
2. What we want in the third column of the data files is ```(f - f(bestparameter))^+ + alpha * sum g_i^+```, not ```(f - f(bestparameter))^+ + alpha * sum g_i^+ - f(bestparameter)```.

nikohansen:
AFAICS, 
1) we keep track of all solutions. Also from _all_ infeasible solutions the trigger value is computed to check whether they trigger a new target.
[...]

PR #1726, opened 2018-03-14
working on logger

asmaatamna:
Implementing point 2 in the first comment on issue #1706


brockho:
Before to review the code (sorry for not having done this earlier), I suggest to also address directly the latest comment in #1706.

PR #1736, opened 2018-04-10
Asma orthogonality check

asmaatamna:
Added a function ```coco_is_orthogonal()``` that checks if the rotation matrix ```M``` used in ```transform_vars_affine.c``` is orthogonal by computing ```MM^T```


nikohansen:
@asmaatamna could you try to fix the conflicts?

nikohansen:
Nice. How about also addressing the comments which had been made on the pull request before the update?

asmaatamna:
@nikohansen I don't see any comment on this pull request. Do you know how to display them?
The only comments I can see are the ones on pull request #1716 (which are not the ones that you added two days ago)

brockho:
I think, I forgot to commit my comments - but clearly that could have not been the comments that you have in mind @nikohansen...

nikohansen:
Comments which come immediately in my mind:
- we would like to have a few more test cases, not only the identity matrix
- code and comment of the orthogonality check function disagree. The code to does not return `1` iff `M*M^T` is the identity.
- `1e-9` should be defined as a constant, such that only one place would need to be changed if the parameter changes.

See also April 19 [here](https://gitlab.inria.fr/randopt/team/wikis/weeklymeeting)

nikohansen:
Find/search `1e-9` is your friend, for some reason I couldn't make a comment online.

nikohansen:
Just found this related issue #1759

nikohansen:
and this #1757

nikohansen:
Just a reminder, mark the request as WIP while you are (still) working on it or close it and reopen this or a new when it is ready.

code-experiments/test/unit-test/test_coco_utilities.c
```diff
@@ -444,11 +444,30 @@ static void test_coco_is_orthogonal(void **state) {
...
   assert(coco_is_orthogonal(M, 2, 2));
+
+  /* Case of a non-orthogonal square matrix */
+  M[0] = 1.;
+  M[1] = 1.;
+  M[2] = 0.;
+  M[3] = 1.;
+  assert(!coco_is_orthogonal(M, 2, 2));
+
+  /* Case of an orthogonal diagonal matrix (almost identity matrix)
```
nikohansen: I don't see that 1e-9 is almost one. To me it looks very different from one.

code-experiments/test/unit-test/test_coco_utilities.c
```diff
@@ -444,11 +444,30 @@ static void test_coco_is_orthogonal(void **state) {
...
+  /* Case of an orthogonal diagonal matrix (identity matrix) */
   M[0] = 1.;
   M[1] = 0.;
   M[2] = 0.;
   M[3] = 1.;
   assert(coco_is_orthogonal(M, 2, 2));
+
+  /* Case of a non-orthogonal square matrix */
+  M[0] = 1.;
+  M[1] = 1.;
```
nikohansen: why not `0.01`? Tests are supposed to stress the code, not to make it easy to pass.

asmaatamna:
The ```transform_vars_affine(): rotation matrix is not orthogonal``` warning was triggered by the following functions (case of ```bbob``` and ```bbob-biobj```):
- attractive sector
- sharp ridge
- Rastrigin
- Weierstrass
- Schaffers
- Katsuura
which is consistent with the definitions (introducing the diagonal matrix ```\lambda^\alpha``` breaks orthogonality of affine transformations)

code-experiments/test/unit-test/test_coco_utilities.c
```diff
@@ -436,6 +436,42 @@ static void test_coco_is_inf(void **state) {
...
+
+  /* Case of an orthogonal diagonal matrix (identity matrix) */
+  M[0] = 1.;
+  M[1] = 0.;
+  M[2] = 0.;
+  M[3] = 1.;
+  assert(coco_is_orthogonal(M, 2, 2));
+
+  /* Case of a non-orthogonal square matrix */
+  M[0] = 0.01;
```
nikohansen: Should be rather ` M[0] = 1.0;` ?

PR #1742, opened 2018-04-18
scala support

pszynk:
This adds support for the Scala language. It reuses most of the Java code.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -75,8 +75,16 @@ jobs:
             apt-get install -y openjdk-8-jdk
             updatedb
       - run:
-          name: Run coco octave tests
+          name: Run coco java tests
           command: python3 do.py test-java
+          no_output_timeout: 1800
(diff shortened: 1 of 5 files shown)


brockho:
Moving the pull request to the `islandblue-b301a2d` branch allowed us to check that all tests run through on Jenkins (also on Windows). But the non-passing test on Appveyor remain.

I tried one change when creating the paths (last commit in this branch), but it did not solve (yet) the problem. Any ideas?

brockho:
Thanks @pszynk for checking this. I have myself also no experience with scala (and no recent one with java either), but I would have a Windows laptop where I could try to run some tests if you tell me what I shall do/install/run. But it will be of course easier if you can continue to just use the AppVeyor testing environment instead.

brockho:
This is a very, very old PR and right now shows conflicts in three files. I'll close this one and if needed, it should be opened again.

PR #1779, opened 2018-05-11
Replacing cmocka with minunit

ttusar:
Replaced cmocka with a C89-compliant implementation of MinUnit (some of MinUnit's functionalities like timing are lost, but what we actually need is there).


ttusar:
You are right, sorry. I have created a new pull request (see #1785), so I'm closing this one.

PR #1785, opened 2018-05-17
Replacing cmocka with minunit (using a different branch)

ttusar:


ttusar:
You are right, thanks for this. I saw these warnings at some point, but then forgot to fix them...

ttusar:
As it turns out, the linux test server is even more pedantic, so I have done some additional changes to take care of warnings due to shadowed declarations. 
Now we should be warning-free...

PR #1787, opened 2018-05-17
Removed cmocka from makefile

ttusar:


nikohansen:
@ttusar , merging into the `devel-test#` branches is also a dangerous habit, because you have (by definition) no control over what is in the branch at the time of the merge. The procedure we agreed upon is to delete the branch and then recreated it with a push.

To get the current branch into the test branch:
```
git push coco :devel-test3  # deletes the branch
git push coco HEAD:devel-test3  # pushes the current HEAD
```

ttusar:
OK

PR #1809, opened 2018-06-19
Computing unconstrained optimum for constrained f_rastrigin

asmaatamna:
We need the optimum to be on the integer lattice in this case.


nikohansen:
I don't think you want to merge into `master`, do you?

PR #1840, opened 2019-01-29
Kostas ls removing coco random new

kostasvar:
It removes the usage of `coco_random.c` for the random permutations, for the block rotations as well as for the definition of the Gallagher and Lunacek Bi-Rastrigin functions.

### DIFF
--- a/code-experiments/src/f_lunacek_bi_rastrigin.c
+++ b/code-experiments/src/f_lunacek_bi_rastrigin.c
@@ -211,7 +211,7 @@ static void f_lunacek_bi_rastrigin_versatile_data_free(coco_problem_t *problem) 
  * @brief Uses the core function to evaluate the sub problem.
  */
 static void f_lunacek_bi_rastrigin_sub_evaluate_core(coco_problem_t *problem, const double *x, double *y) {
-  
+
   assert(problem->number_of_objectives == 1);
   y[0] = f_sphere_raw(x, problem->number_of_variables);
 }
@@ -221,7 +221,7 @@ static void f_lunacek_bi_rastrigin_sub_evaluate_core(coco_problem_t *problem, co
  * @brief Allocates the basic lunacek_bi_rastrigin sub problem.
  */
 static coco_problem_t *f_lunacek_bi_rastrigin_sub_problem_allocate(const size_t number_of_variables) {
-  
+
   coco_problem_t *problem_i = coco_problem_allocate_from_scalars("lunacek_bi_rastrigin_sub function",
                                                                  f_lunacek_bi_rastrigin_sub_evaluate_core, NULL, number_of_variables, -5.0, 5.0, 0.0);
   problem_i->versatile_data = NULL;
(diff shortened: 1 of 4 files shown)


code-experiments/src/f_lunacek_bi_rastrigin.c
```diff
@@ -357,9 +357,11 @@ static coco_problem_t *f_lunacek_bi_rastrigin_permblockdiag_bbob_problem_allocat
   ((f_lunacek_bi_rastrigin_versatile_data_t *) problem->versatile_data)->sub_problem_mu1 = f_lunacek_bi_rastrigin_sub_problem_allocate(dimension);
 
   /* set sign_vector */
```
brockho: looks like an unnecessary comment here

code-experiments/src/transform_vars_blockrotation_helpers.c
```diff
@@ -90,27 +90,30 @@ static void coco_compute_blockrotation(double **B, long seed, size_t n, size_t *
   double **current_block;
   size_t i, j, k; /* Loop over pairs of column vectors. */
   size_t idx_block, current_blocksize,cumsum_prev_block_sizes, sum_block_sizes;
-  size_t nb_entries;
-  coco_random_state_t *rng = coco_random_new((uint32_t) seed);
-  
+  size_t nb_entries, current_nb_entries;
```
brockho: Better:
```C
size_t nb_entries;
size_t current_nb_entries;
```
because two declarations in one line are, IIRC, not allowed in the C89 ("ANSI C") standard.

code-experiments/src/transform_vars_permutation_helpers.c
```diff
@@ -53,10 +56,26 @@ static void coco_compute_random_permutation(size_t *P, long seed, size_t n) {
 static long coco_random_unif_int(long lower_bound, long upper_bound, coco_random_state_t *rng){
   long range;
   range = upper_bound - lower_bound + 1;
+  printf("%ld\n", ((long)(coco_random_uniform(rng) * (double) range)) + lower_bound);
```
brockho: Looks like a leftover from testing and should be removed, shouldn't it?

code-experiments/src/transform_vars_permutation_helpers.c
```diff
@@ -53,10 +56,26 @@ static void coco_compute_random_permutation(size_t *P, long seed, size_t n) {
...
+  long range, rand_int;
+  double *tmp_uniform;
+  tmp_uniform=coco_allocate_vector(1);
+  bbob2009_unif(tmp_uniform, 1, seed);
+  range = upper_bound - lower_bound + 1;
+  rand_int = ((long)(tmp_uniform[0] * (double) range)) + lower_bound;
+  coco_free_memory(tmp_uniform);
+  return rand_int;
+}
+
```
brockho: This function should probably be moved to a more prominent place because this is more general than just being a helper function of `transform_vars_permutation`. The question is: where shall it go? Most likely to `coco_random.c`, I guess?

code-experiments/src/transform_vars_blockrotation_helpers.c
nikohansen: It's bad, but I don't think C89 is that bad. The restriction is that variables can only be defined at the beginning of a scope. Declaring several variables at once is IMHO reasonable coding style.

PR #1848, opened 2019-02-15
Rw devel mixint ls merge

ttusar:


nikohansen:
Just FTR (not sure if this applies), there is some risk in merging into the devel-test branches, as their state is not guarantied. A correct sequence to test the current HEAD would be:
```
git push -f coco :devel-test3  # delete remote test branch
git push coco HEAD:devel-test3
```

ttusar:
Thanks for reminding me (again) about the correct procedure.

ttusar:
Oh, just realized my comment above might look a bit snarky - it really wasn't meant to be. The "again" was for me, because I haven't done this in a long time and actually need reminding, although I should have remembered...

nikohansen:
No worries!

PR #1860, opened 2019-03-11
Large-scale and mixed-integer suites

ttusar:
Merged from development


ttusar:
@kostasvar Can you produce the new data for the regression tests?

ttusar:
Does the constrained test suite count as a fully supported one? (I'm asking because of the documentation within the code)

code-experiments/build/matlab/my_optimizer.m
```diff
@@ -1,9 +1,13 @@
-function my_optimizer (problem, lower_bounds, upper_bounds, budget)
+function my_optimizer (problem, lower_bounds, upper_bounds, num_integer_vars, num_constraints, budget)
 n = length(lower_bounds);
 delta = upper_bounds - lower_bounds;
 for i= 1:budget
     x = lower_bounds + rand(1,n) .* delta;
-    if cocoProblemGetNumberOfConstraints(problem) > 0
+    % Round the variable values that need to be integer
```
ttusar: Is there a better way to explicitly show that problems can also be mixed-integer (this code makes little sense)?

brockho:
Since there was not much progress on the constrained test suite since the last release (where the `bbob-constrained` suite was not supported), I think that the constrained test suite should also not be counted as a fully supported one for the next 2.3 release. We have reserved the drafted release 2.4 for the constrained test suite.

ttusar:
@brockho okay, thanks

code-experiments/build/matlab/my_optimizer.m
kostasvar: > @kostasvar Can you produce the new data for the regression tests?

They are uploaded now.

kostasvar:
The regression data are uploaded now.

ttusar:
@kostasvar Have you produced regression data also for the two mixed-integer suites?

ttusar:
@kostasvar Nevermind, I see that you have, otherwise there would be more fails. 

So, it seems that things *mostly* work - the regression test fails for the mixed-integer suite on Linux. I will try to find the problem.

ttusar:
All tests are finally clear \o/

Can we merge this pull request now?

brockho:
I would say yes.

ttusar:
@nikohansen Do you agree?

PR #1882, opened 2019-04-12
Postprocessing bbob-biobj-ext data

brockho:
together with bbob-biobj data by removing all functions with ids larger than 55 if both test suites are detected.

[tests are currently running on `devel-test3`]


brockho:
The branch `brockho:pp-bbobbiobjext` needs to be cleaned up before a merge. I therefore close this pull request to avoid an unwanted merge.

PR #1913, opened 2019-08-02
Rw communication

ttusar:
Implementation of external evaluation through sockets


ttusar:
@nikohansen and @brockho, I have implemented external evaluation using socket communication, and two new suites, `toy-socket` and `toy-socket-biobj`, to demonstrate how it works. The examples of servers (external evaluators) are done in C and Python. AFAIKS, all* combinations of platforms (Win, Linux, Mac) and languages (C, Python, Java, Matlab) work with both of them.

I did this because we need it for the GBEA suites, but I imagine it should come in handy in general, especially of course for including new real-world suites into COCO. See my `rw-communication` [fork](https://github.com/ttusar/coco/tree/rw-communication) and [this README](https://github.com/ttusar/coco/blob/rw-communication/code-experiments/rw-problems/toy-socket/README.md) if you want to try it yourselves.

I came to a stop when trying to automatically compile and run experiments on these two suites through `do.py`. The problem is that the server needs to be ran in a new (additional) terminal and I do not know how to do this. Do you know how this could be achieved (cross-platform)?

Anyway, any comments/ideas regarding this new code are welcome.

*I haven't tried Matlab on Linux and Mac, because I don't have it available... and I haven't tried Octave at all.

ttusar:
@brockho Says it should be possible to run the server process in the background

nikohansen:
> Also, can these two suites be omitted from the regression tests (but not from python experiments)? How?

Late to the party, I guess it would make sense to introduce an exclusion list for this purpose?

ttusar:
This is a really old and outdated pull request. If we want to add this functionality in the future, we will need to start from a fresh pull request.

PR #1929, opened 2020-01-06
addressing issue 1923 (also for coco-docs paper)

brockho:
should be merged as soon as tests run through on Jenkins


code-postprocessing/cocopp/compall/pprldmany.py
```diff
@@ -877,7 +880,7 @@ def formatlabel(val, pos):
                           # Relative additional space numbers are
                           # bottom, left, 1 - top, and 1 - right.
                           # bottom=0.13 still clips g in the log(#evals) xlabel
-                          subplots_adjust=dict(bottom=0.135, right=0.735),
+                          subplots_adjust=pprldmany_adjust,
```
nikohansen: Maybe this should read
```python
subplots_adjust=dict(bottom=0.135,
                     right=0.735,
                     top=0.92 if len(dictFunc) == 1 else 0.99)
```
and the `pprldmany_adjust` variable and the above code duplication should go away.

Issue #1968, opened 2020-05-22
A recipe for running experiments with different dimensions


nikohansen:
Re. https://github.com/numbbo/coco/issues/1966#issuecomment-629847985

Shall we have a recipe for this, one that is not deeply buried in a soon-to-be-closed issue?

It could go into the [howto folder](https://github.com/numbbo/coco/tree/master/howtos), I guess basically a copy-paste of the above linked comment?

@ttusar ?

ttusar:
See https://github.com/numbbo/coco/pull/1969 

I think you would probably be able to add some more useful information under the "disclaimer" part (such as that if dimensions go over some number, one would be better off using the large-scale suite instead of the bbob - at least I imagine this would be the way to go...)

ttusar:
It's now available on the development branch.

PR #1969, opened 2020-05-22
Addressing #1968

ttusar:
- Added a "how to" for experimenting with different dimensions
- Computed the number of dimensions from the size of the array
- Cleaned up the compiler warnings


ttusar:
Other than the failed leak check on the Mac, everything looks fine

nikohansen:
Are you sure it is OK to merge from a devel-test branch? I was in the believe that they are in an arbitrary state and can (and should) be forced overwritten at any point in time?

ttusar:
You are right, I'm closing this one and opening a new pull request (see #1970).

PR #1976, opened 2020-06-02
Merge Development into Master...

brockho:
...to see the effect of the new issue templates [a new release is planned for later]


nikohansen:
It looks like we cannot merge when the leak-check fails.

code-experiments/src/suite_toy.c
```diff
@@ -25,8 +25,9 @@ static coco_suite_t *suite_toy_initialize(void) {
 
   coco_suite_t *suite;
   const size_t dimensions[] = { 2, 3, 5, 10, 20 };
+  const size_t num_dimensions = sizeof(dimensions) / sizeof(dimensions[0]);
 
-  suite = coco_suite_allocate("toy", 6, 3, dimensions, "instances: 1");
+  suite = coco_suite_allocate("toy", 6, num_dimensions, dimensions, "instances: 1");
```
nikohansen, line 30: @ttusar are you sure that this change is correct? Did the toy suite change?

nikohansen: OK, I guess I see now that this just changes the suite to contain different dimensions.

PR #1991, opened 2020-09-23
Silence matplotlib 3.3.2 warning

MLopez-Ibanez:
UserWarning: FixedFormatter should only be used together with FixedLocator

by setting the xticks before their labels.


brockho (approved):
Looks reasonable and, at first sight, should not have an impact on the output of the postprocessing.

PR #2018, opened 2021-08-24
Devel test2

ttusar:


ttusar:
The latest issue doesn't have to do with any changes I made, but seems to be an incompatibility issue between our version of Cython and Python 3.9. The error message is `error: ‘PyTypeObject’ {aka ‘struct _typeobject | | ’} has no member named ‘tp_print’` and here is some more information: https://stackoverflow.com/questions/65462530/pyclipper-installation-error-tp-print-is-not-a-member-of-typeobject

It seems we need to upgrade our Cython version, but I don't know how to do that...

PR #2056, opened 2021-12-03
Constrained test suite with 54 functions

nikohansen:
As #2052 

This PR to merge into the current development branch the last features we developed for the constrained test suite:
- Inactive constraints
- Active and inactive constraints indices are randomly shuffled
- The first constraint is modified to avoid trivial Lagrange multipliers
- The rotated Rastrigin is added to the list of test problems
- Unofficial supplementary suites
   - `bbob-constrained-no-disguise` (where the Lagrange multipliers are `[1, 0, ..., 0]`)
   -  `bbob-constrained-active-only` which describes itself


nikohansen:
@paulduf could you check the last four commits?

paulduf:
Ok for me, you can merge

PR #2070, opened 2022-02-14
Experiment: Lots of small C code fixes

olafmersmann:
Lots of small code fixes to catch more edge cases, remove dead code.

### DIFF
--- a/code-experiments/src/coco.h
+++ b/code-experiments/src/coco.h
@@ -11,6 +11,48 @@
 
 #include <stddef.h>
 
+/* Macro to tag function declarations:
+ *
+ * COCO_NORETURN:
+ *   Function never returns. Mainly useful for linters and static analysis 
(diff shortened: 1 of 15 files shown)


code-experiments/src/coco.h
```diff
@@ -465,7 +465,7 @@ void coco_free_memory(void *data);
 /**
  * @brief Signals a fatal error.
  */
-void coco_error(const char *message, ...);
+void coco_error(const char *message, ...) __attribute__((noreturn)) __attribute__((__noreturn__));
```
paulduf, line 468: This seems to be the culprit for the AppVeyor failure

olafmersmann: I've hidden the attributes behind macros which are defined based on the used compiler. Let's hope this fixes it.

PR #2071, opened 2022-02-18
Experiment: Add example CMake and meson build files

olafmersmann:
CMake and meson are build systems that have both gained popularity. This PR adds example build files for both so that users might use them as starting points if their project already uses CMake or meson. It should also be a good option for users on Windows where CMake can generate Visual Studio project files.

### DIFF
--- a/code-experiments/build/c/meson.build
+++ b/code-experiments/build/c/meson.build
@@ -1,0 +1,18 @@
+project('example_experiment', 'c',
+  default_options: ['warning_level=3', 'buildtype=release']
+  )
+
+cc = meson.get_compiler('c')
+m_dep = cc.find_library('m', required : false)
+
+coco_lib = static_library('coco', 
+  sources: 'coco.c',
+  dependencies: m_dep
+  )
+
+executable('example_experiment', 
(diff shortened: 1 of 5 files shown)


olafmersmann:
Should these be checked in CI?

code-experiments/build/c/meson.build
```diff
@@ -0,0 +1,20 @@
+project('example_experiment',
+  default_options: ['warning_level=3', 'buildtype=release']
+  )
+
+add_languages('c')
```
eli-schwartz, line 5: The idiomatic way to add a language is in the project function.

```meson
project('example_experiment', 'c', default_options: ['warning_level=3'])
```

(Hmm, why isn't the cmake build also setting `-Wall -Wextra -Wpedantic`?)

olafmersmann: Thanks for the feedback! Changed the language specification to the more idiomatic way. Also tried to add warnings to the CMakeLists.txt, but afaik there's no easy and compiler agnostic way to do this.

nikohansen (approved):

PR #2074, opened 2022-03-02
Add the latex template

paulduf:
- Correct captions for the constrained case where it was forgotten
- LaTeX template for experiments on the constrained testbed, for single- and many- algorithm(s)


paulduf:
The test failed with the following error: `Build execution time has reached the maximum allowed time for your plan (60 minutes).`

nikohansen:
Part of the problem may be that the Python 2 installation takes too much time?

PR #2078, opened 2022-03-03
Release 3.0

paulduf:
Triggering all tests before merging to master


nikohansen:
Can we check the test status on a branch like we can see it with a pull request?

PR #2088, opened 2022-03-12
Release 2.6.1

paulduf:
Correct an undesirable shift of the flat region  of the linear slope function in the constrained testbed, using an unused-but-already-implemented argument  in `transform_vars_shift`.


brockho (approved):
Looks fine. All tests pass. Thus, we should do the release.

PR #2118, opened 2022-05-23
Add `.gitignore` files for c and python.

luleyleo:
This covers the following commands:
- `build-c`
- `run-c`
- `test-c`
- `build-python`
- `build-rust`
- `run-rust`
- `test-rust`

It does not cover:
- `run-python`
- `*-java`
- `*-matlab`

This is only due to me not having these setup on my computer, but they would be easy to add for someone who has them.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,6 @@ REVISION
 # build files
 __pycache__
 *.o
+
+# tests
+code-experiments/test/example-test
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,6 @@
+VERSION
+REVISION
+
+# build files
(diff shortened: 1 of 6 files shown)


nikohansen:
I am not sure I understand why do we want to ignore the text and setup files under [`code-experiments/build/python`](https://github.com/numbbo/coco/pull/2118/files#diff-19119f3942cd05d25a5d96686a4603242296759aad8fb1b1531f4e9abd8f3707)?

luleyleo:
@nikohansen can we merge this?

luleyleo:
@nikohansen can we merge this?

PR #2119, opened 2022-05-24
Update Rust bindings.

luleyleo:
This has three changes:
- Added missing `Suite::problem_by_function_dimension_instance` function
- Prevent `Problem` from outliving its `Suite`
- Incorporate internal `C` code changes into the Rust package


luleyleo:
Can we merge this?

PR #2139, opened 2022-09-07
Compute an initial feasible solution

paulduf:
For the constrained test bed, the technique to find a feasible initial solution to the final problem was introducing side effects.
Indeed, a feasible point was found by line search between the optimal solution and the initial solution *which is feasible for the linearly constrained problem before applying a transformation $T$*. If the nonlinear transformation T bends the search space the wrong way, then the line search finds a point very close to the optimum, which is not desirable.
It is easy to show that applying the inverse of $T$ to the initial solution (before shifting by $x^{opt}$) yields a feasible initial solution for the final problem.
This feature is implemented in this PR.

# New feature: Compute the inverse transform of non-linear transformation

## Repair of the initial solution

The goal is to be able to query the inverse of the oscillating and asymmetric transformations.
They are respectively in `transform_vars_asymmetric.c` and `transform_vars_oscillating.c`.
The transformations are strictly monotonous, hence invertible but their inverse does not yield a closed-form (a tool like SymPy is not able to get a closed-form).
There is still hope to compute the inverse numerically using a _root finding_ algorithm like Brent's method.
This method is implemented in `brentq.c` under the `brentq` function.
This function comes from SciPy and the license is updated accordingly.
It takes a `callback_type` first argument which is the mathematical function $T$ for which we want to find a root. 
It is slightly modified to accept also an argument `double y;` such that we search the root of $T(x) - y$.
The `brentq` function is then wrapped into another function, `brentinv`, which features a simple heuristic to find the initial bracket $[a , b]$  and calls `brentq` with these parameters.

We described the procedure to invert a univariate mathematical function.
In the case of the bbob-constrained test suite, we want to invert a multivariate transformation.
This transformation applies coordinate-wise so it falls back to apply this procedure to each coordinate.
Some changes to the existing code were needed to do so.

## Handle solutions out of bounds

If the inverse transformation yields a solution out of bounds, then we have to do a lign search along the curve induced by $T^{-1}$.
This is implemented in each `transform_vars_*.c` file. 

### to the transformation code

*I describe what was done for the asymmetric transformation $T_{asy}  but this applies to the oscillating transformation $T_{osz} as well (where it is even simpler because it does not depend on the dimension).*

First, the asymmetric transformation code was duplicated in `transform_vars_asymmetric_evaluate_function` and in `transform_vars_asymmetric_evaluate_constraint`.
So I extracted the transformation from these into the `static transform_vars_asymmetric_data_t *tasy` function. 
Then, I had to create the univariate transformation `tasy_uv` from the existing `tasy` function.
The signature of this function is static double `tasy_uv(double xi, tasy_data *d)` and `tasy_data` is a new structure containing additional parameters to the transformation.
In particular, the asymmetric transformation depends on the coordinate index `i` and total number of variables `n` so these are two elements of the structure.
This data can be passed to the brentinv method because it signature is `double brentinv(callback_type f, double y, void *func_data)`.

### to the declaration code

The inverse transform needs to be applied to the feasible direction in the same time as we apply the transformation to the problem.
A proposition is then to have a new function calling the `brentinv` method coordinate wise, and to call this function in the method to instantiate each problem in the `suite_cons_bbob_problems.c` file.

For example, in  `static coco_problem_t *f_ellipsoid_c_linear_cons_bbob_problem_allocate`, just after

```c
problem = transform_vars_oscillate(problem);
feasible_direction = transform_inv_feas_dir_asymmetric(problem, feasible_direction)  # this is new
transform_inv_feas_dir_asymmetric(problem, feasible_direction)  # or this (why problem is not changed inplace ?)
```

### Composed transformation

For Rastrigin the transformation is composed, $T = T_{asy} \odot T_{osz}$, so to keep the implementation simple I just created a new `transform_inv_feas_dir_composed` method under the `transform_vars_composed.c` file.

### to the testing code

Tests to check initial solutions are feasible and remain within the bounds are added in the existing file `code-experiments/test/integration-test/test_bbob-constrained.c`


paulduf:
I checked with the boxplots (from a notebook that I should clear and upload somewhere on this repo on another owned by the `numbbo` organization) that:
 for almost all functions we have $f(x^{init}) - f(x^{opt}) > 10^2$ (for a few of them it is below but not much so maybe the first target is solved)
which is what we want

nikohansen (approved):

paulduf:
It remains to handle the exceptions / scenarios where `brentq` do not converge. The original code uses SciPy's specific objects, and I don't really know how you want to handle this in COCO. We can just raise an error in this case, as it should never happen for the -- monotonous -- transformation -- having fixed point -- at 0 we use.

nikohansen:
Raising an error doesn't look like a good option. A warning that the initial solution proposal for the respective instance is not feasible should do.

paulduf:
Here I mean in the inner method `brentq` when the solution did not converge. When this detected, the outer method can not replace the initial solution and use the one without inverse transform, which may not be feasible, and indeed throw a warning.
I'm asking what is the best way to do this in C. A try/except clause look not available inC, so AFAICS we need to add a boolean-like `has_succeed` parameter returned by the inner Brent method to the outer one.

LICENSE
```diff
@@ -52,4 +52,5 @@ We acknowledge in particular:
   ``code-experiments/test/unit-test/minunit_c89.h``, distributed under
   the terms of the MIT Licence.
 * the Rust code by Leopold Luley in `code-experiments/build/rust/`, distributed unter the terms of both the MIT License and the Apache 2.0 License.
+* the code for Brent's method in ``code-experiments/src/brentq.c`` is a modified copy of ``scipy.optimize.brentq`` originally written by Charles Harris and distributed under the BSD 3-Clause "New" or "Revised" License.
```
nikohansen, line 55: Don't we need the license file here in order to respect the license?

code-experiments/src/brentq.c
```diff
@@ -0,0 +1,195 @@
...
+
+            fpre = fcur;
+            fcur = fblk;
+            fblk = fpre;
+        }
+
+        delta = (xtol + rtol*fabs(xcur))/2;
+        sbis = (xblk - xcur)/2;
+        if (fcur == 0 || fabs(sbis) < delta) {
+            /* TODO: converged*/
```
nikohansen, line 103: Fix todo?

code-experiments/src/transform_vars_asymmetric.c
```diff
@@ -115,31 +154,66 @@ static coco_problem_t *transform_vars_asymmetric(coco_problem_t *inner_problem,
...
-  }
   
   if (inner_problem->number_of_objectives > 0 && coco_problem_best_parameter_not_zero(inner_problem)) {
     coco_warning("transform_vars_asymmetric(): 'best_parameter' not updated, set to NAN");
     coco_vector_set_to_nan(inner_problem->best_parameter, inner_problem->number_of_variables);
   }
   return problem;
 }
+
+static void transform_inv_feas_dir_asymmetric(coco_problem_t *problem, const double *xopt) {
```
nikohansen, line 167: Docstring is missing. We should have at least broad idea what this function is for.

code-experiments/src/brentq.c
paulduf: Yes, just forgotten, nothing special to do here

code-experiments/test/unit-test/test_brentq.c
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * Tests the function coco_observer_targets_trigger.
+ */
+
+MU_TEST(test_brentq) {
+    double y;
+
+    y = brentq((callback_type) &sqrt, 4, 1, 4, 1E-14, 1E-10, 200, NULL);
+    mu_check(y - 2 < 1E-14);
```
nikohansen, line 13: Don't we also want to check that `2 - y < 1E14`?

code-postprocessing/latex-templates/templatepreamble.tex
```diff
@@ -65,7 +65,6 @@
   \begin{rotate}{90}\hspace{#1em}#2
   \end{rotate}}
 \newcommand{\DIM}{\ensuremath{\mathrm{DIM}}}
-\newcommand{\ERT}{\ensuremath{\mathrm{ERT}}}
```
nikohansen, line 68: Why is \ERT is not needed anymore?

paulduf: This was just a mistake I committed without notice, fixed

nikohansen (approved):

PR #2154, opened 2023-01-18
Reformat C code to conform  to our style guide.

olafmersmann:
Clean up the formatting of all C code.
Ideally we would want to automate this on commit with something like [pre-commit](https://pre-commit.com/) to ensure we don't regress.


olafmersmann:
This should *not* merge until the new linear logger by @ttusar is merged.

PR #2159, opened 2023-01-22
Added ExampleNonAnytime in Java - RS Algorithm

ElieTrigano:


code-experiments/build/java/ExampleExperimentNonAnytime.java
```diff
@@ -0,0 +1,233 @@
...
+                    /* Call the optimization algorithm for the remaining number of evaluations */
+                    MyRandomSearch(evaluateFunction,
+                            dimension,
+                            PROBLEM.getNumberOfObjectives(),
+                            PROBLEM.getNumberOfConstraints(),
+                            PROBLEM.getSmallestValuesOfInterest(),
+                            PROBLEM.getLargestValuesOfInterest(),
+                            evaluationsRemaining,
+                            randomGenerator);
+
```
brockho, line 137: What is missing here is the reset of the problem in order to prepare logging for the next budget like in the non-anytime [python code](https://github.com/numbbo/coco/blob/master/code-experiments/build/python/example_experiment_non_anytime.py#L110), or am I wrong?

brockho (changes requested):

brockho:
Dear Elie,

We also see that there is an error on CircleCI due to a non-applied fix (which has been done in the development branch but not yet in the master branch). You based your pull request on the master branch and not on development and thus missed the fix. So would you be so kind and create yet another pull request, now from the development branch which contains the fix for the CircleCI? Thanks in advance.

ElieTrigano:
Dear Dimo,Sorry again for all these issues, I’m not used to work with GitHub for now so I make mistakes. I’ll try to make this changes asap!Best regards,ElieOn 24 Jan 2023, at 14:56, Dimo Brockhoff ***@***.***> wrote:﻿
Dear Elie,
We also see that there is an error on CircleCI due to a non-applied fix (which has been done in the development branch but not yet in the master branch). You based your pull request on the master branch and not on development and thus missed the fix. So would you be so kind and create yet another pull request, now from the development branch which contains the fix for the CircleCI? Thanks in advance.

PR #2211, opened 2023-08-09
First PR from the forked repository to the dev-bbob-noisy branch of the numbbo/coco

FMGS666:
I started working on the implementation of the bbob-noisy test suite. There is already a good amount of changes (hopefully, they make sense, but we'll surely discuss about it very soon), so I thought it could be a good idea to make the fist PR. Professor Brockhoff added me to the organization, saying that committing directly to the numbbo/coco development branches makes it easier to test the modification as it is automated within github.


brockho:
It is difficult to see because the CircleCI fails with some COCO-unrelated error, but there is definitely one test on your code that does not pass: 
```
COCO FATAL ERROR: coco_suite_get_instances_by_year(): suite 'bbob-noisy' has no years defined
Command exited with code -1
```

Otherwise, we can (and I think you can do yourself) accept the PR. You could then also work directly on this new branch to reduce the amount of PRs (simple commits are okay if they are handled directly within `dev-bbob-noisy`).

PR #2214, opened 2023-08-29
Devel test2

ttusar:
Added some functions to the Java interface to enable the usage of COCO problems separately (not within a benchmarking experiment).

### DIFF
--- a/code-experiments/build/java/CocoJNI.c
+++ b/code-experiments/build/java/CocoJNI.c
@@ -188,6 +188,29 @@ JNIEXPORT void JNICALL Java_CocoJNI_cocoFinalizeSuite
 
 /*
  * Class:     CocoJNI
+ * Method:    cocoSuiteGetNumberOfProblems
+ * Signature: (J)J
+ */
+JNIEXPORT jlong JNICALL Java_CocoJNI_cocoSuiteGetNumberOfProblems
+(JNIEnv *jenv, jclass interface_cls, jlong jsuite_pointer) {
+
+  coco_suite_t *suite = NULL;
+  jlong jresult;
+
+  /* This test is both to prevent warning because interface_cls was not used and to check for exceptions */
+  if (interface_cls == NULL) {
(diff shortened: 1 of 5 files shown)


ttusar:
The test `cargo install bindgen-cli` fails with error: 
````
error: failed to compile `bindgen-cli v0.66.1`, intermediate artifacts can be found at `/tmp/cargo-installSw5dGF`

Caused by:
  package `clap v4.4.1` cannot be built because it requires rustc 1.70.0 or newer, while the currently active rustc version is 1.67.1
  Try re-running cargo install with `--locked`

Exited with code exit status 101
````

Help!

ttusar:
Additionally, the `Coco-test2-win/dotype=verify-postprocessing,node=numbbo-win-mingw` test fails as follows:

````
Validation failed! File latex_commands_for_html.html should be regenerated!
Original line:                         an <span class="roman">ERT</span>&nbsp;of !!PPFIGS&#8722;FTARGET!!&times;<span class="roman">DIM</span>. Different symbols correspond to different algorithms given in the legend of f<sub>1</sub> and f<sub>24</sub>. Light symbols give the maximum number of evaluations from the longest trial divided by dimension. Black stars indicate a statistically better result compared to all other algorithms with p &lt; 0.01 and Bonferroni correction number of dimensions (six).  

Generated line:                         an <span class="roman">ERT</span>&nbsp;of !!PPFIGS&#8722;FTARGET!!&times;<span class="roman">DIM</span>. Different symbols correspond to different algorithms given in the legend of f<sub>1</sub> and f<sub>24</sub>. Light symbols give the maximum number of evaluations from the longest trial divided by dimension. Black stars (if present) indicate a statistically better result compared to all other algorithms with p &lt; 0.01 and Bonferroni correction number of dimensions (six).  
````

As far as I can see, `(if present) ` was added to the text. I have no idea what it means to regenerate the file `latex_commands_for_html.html`, therefore, help would be appreciated.

PR #2253, opened 2023-12-08
Try fixing an Octave installation problem

brockho:


brockho:
This did not work. Maybe moving the tests to python 3.9 (from 3.8) will remove the bug we see.

PR #2295, opened 2024-06-07
Fixed logger filenames

ttusar:
Fixed issue with repeated file names in a single info file + increased the limit for unique files/folders.


nikohansen:
As far as I can see so far, it seems to work perfectly fine.

PR #2336, opened 2024-11-13
This was an error in the implementation

JJ:
Captura de pantalla de 2024-11-13 17-34-02
Besides being equivalent to 1 that way, you need to go no further than the formula (above). It is now equivalent to D^2; before it was simply 1.

### DIFF
--- a/code-experiments/src/f_katsuura.c
+++ b/code-experiments/src/f_katsuura.c
@@ -43,7 +43,7 @@ static double f_katsuura_raw(const double *x, const size_t number_of_variables) 
   }
   /*result = 10. / ((double) number_of_variables) / ((double) number_of_variables)
       * (-1. + pow(result, 10. / pow((double) number_of_variables, 1.2)));*/
-  result = 10. / ((double) number_of_variables) / ((double) number_of_variables)
+  result = 10. / ((double) number_of_variables) * ((double) number_of_variables)
   * (-1. + result);
 
   return result;


code-experiments/src/f_katsuura.c
```diff
@@ -43,7 +43,7 @@ static double f_katsuura_raw(const double *x, const size_t number_of_variables)
   }
   /*result = 10. / ((double) number_of_variables) / ((double) number_of_variables)
       * (-1. + pow(result, 10. / pow((double) number_of_variables, 1.2)));*/
-  result = 10. / ((double) number_of_variables) / ((double) number_of_variables)
+  result = 10. / ((double) number_of_variables) * ((double) number_of_variables)
```
nikohansen, line 46: this line can be reduced to
```C
result = 10.
```
because the operations `/ ((double) number_of_variables)` and `* ((double) number_of_variables)` cancel each other out. This lets me believe the original implementation was correct, no?

JJ: Not really, it's 1/D^2, as in the formula. The original implementation was simply 1 (in a roundabout way, because looking at code generated it actually did insert the division of D/D)

JJ (commented):
Answering the question...  *Before* it could be reduced to what @nikohansen said. *Now* it does not.

code-experiments/src/f_katsuura.c
JJ: OK, I get it now. Operator associativity actually makes 10/20/20 be equivalent to 10/(20*20) and 10/20*20 be equivalent to (10/20)*20. What I have done is to use parentheses to make the implementation closer to the formula, and prevent people from applying operator precedence and associativity rules, thus reducing cognitive load.

